PR #11, opened 2014-12-26
Added ClearRegisteredTemplates method. Use it while in development to se...

nkwinder:
Added a method, for clearing all the registered templates. Seems to be working... as expected. Any ideas?

### DIFF
--- a/source/Handlebars/Handlebars.cs
+++ b/source/Handlebars/Handlebars.cs
@@ -32,11 +32,6 @@ namespace Handlebars
             _singleton.RegisterTemplate(templateName, template);
         }
 
-        public static void ClearRegisteredTemplates()
-        {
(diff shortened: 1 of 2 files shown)


rexm:
Seems like that is a more consistent approach.

nkwinder:
New commits revert the initial changes and make the partials overwrite each time they are registered. 

So now in my app, in production, the registerTemplate will be called only in Application Start, but now i have also the option to call registerTemplate in Begin Request, (if in development mode), and changes will be shown without an app restart.

Seems legit? 

p.s What about renaming the function to registerPartial ?

nkwinder:
Any update on this one? Am i missing something?

rexm:
1.1.2 pushed to NuGet

PR #23, opened 2015-01-21
Implemented support for passing properties into a partial.

leethomascook:


leethomascook:
Actually, I left some test code in there - on way home, will re submit shortly

PR #28, opened 2015-02-09
Added helperName to the HandlebarsRuntimeException message

csharpsi:
More detailed error message here would help debug issues.


rexm:
Updated NuGet 1.2.3

PR #35, opened 2015-02-27
Supporting dictionary and accessing with dot notation.

ondrejtomcik:
Using dictionary is common and usefull for embedded object.
https://tools.ietf.org/html/draft-kelly-json-hal-06#section-4.1.2

You can extend your transfer model with embedded optional objects.


rexm:
Updated NuGet to 1.2.7

PR #37, opened 2015-03-05
Implement iterating over generic dictionaries and dynamic objects.

kll:
Only dynamic objects that implement IDynamicMetaObjectProvider can be iterated.
I believe any generic dictionary should work, but I only wrote tests for Dictionary<string, object>.


rexm:
Nice, thanks!

rexm:
Updated NuGet to 1.2.9

Issue #33, opened 2015-02-12
Navigating an array by index doesn't work same as in handlebarsjs


johnrobertallan:
In javascript I can navigate an array using the following syntax

array.1.title

However I am yet to see this work in the .net library. Perhaps I have implemented something incorrectly but it seems straightforward. Any insight would be appreciated.

![handlebars-array-navigation](https://cloud.githubusercontent.com/assets/5693857/6177481/126f5b32-b2bc-11e4-8025-45e31ddc6a5a.png)

rexm:
I missed this particular type of pathing... it'll need to be added.

rexm:
This was fixed by @kll's PR #40 and included in the 1.2.10 NuGet release.

PR #40, opened 2015-03-16
Allow indexing into arrays with paths.

kll:
This fixes issue #33.

This is slightly more flexible than handlebarjs is. Handlebarsjs supports indexes with and without square brackets when a child path is present, but evidently requires the square brackets when there is no child path. See http://stackoverflow.com/questions/8044219/handlebars-access-array-item

I did not immediately see how to easily replicate the exact behavior of handlebarsjs, and it was very easy to support missing square brackets in both cases. So I simply did not include a unit test to test the case that handlebarsjs does not support.

### DIFF
--- a/source/Handlebars.Test/BasicIntegrationTests.cs
+++ b/source/Handlebars.Test/BasicIntegrationTests.cs
@@ -34,6 +34,45 @@ namespace Handlebars.Test
         }
 
         [Test]
+        public void BasicPathArray()
+        {
+            var source = "Hello, {{ names.[1] }}!";
+            var template = Handlebars.Compile(source);
+            var data = new
+            {
+                names = new[] {"Foo", "Handlebars.Net"}
+            };
+            var result = template(data);
+            Assert.AreEqual("Hello, Handlebars.Net!", result);
+        }
+
+        [Test]
+        public void BasicPathArrayChildPath()
+        {
+            var source = "Hello, {{ names.[1].name }}!";
+            var template = Handlebars.Compile(source);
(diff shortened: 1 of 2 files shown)


rexm:
Nice! One minor point for consistency: the AccessMember method is where the existing logic branches based on the instanceType, so it'd be good to have this there as well.

kll:
Ok, it's your library. :) I don't think the names really fit now though. Its accessing an element with an index rather than a member with a name!

rexm:
Closer at least.... we can come up with a different word than "member" :)

rexm:
It may be a day or two before this makes it to NuGet... I am on holiday and did the merge from my phone on the beach :)

rexm:
Updated NuGet 1.2.10

rexm:
For what it's worth (which isn't much, but I found it interesting enough to share!) - I went hunting for a "more correct" name than `AccessMember`, and it turns out that the term "Member" in c# [does include indexers](https://msdn.microsoft.com/en-us/library/ms173113.aspx).

kll:
Yeah, I suppose I was thinking more about the end result which in this case is what is returned from the indexer which I think of as an element. Technically though you are right and it is the indexer itself that is being "accessed" by this code block.

Mostly I was just in a rush to get the PR sent and didn't think about it enough. :) Which reminds me, thanks for going ahead and pushing out a new package from the beach, because I had another PR to file for another project that depended on this one that I needed accepted before I could finally get back to my regularly scheduled development!

PR #45, opened 2015-04-02
Reserve dynamic iteration logic for non-list IDynamicMetaObjectProviders

tsliang:
Some objects implement both IEnumerable and IDynamicMetaObjectProviders, but do
not have dynamic member names (chief example being Newtonsoft.Json JArrays).
This patch ensures that Handlebars.Net only uses dynamic iteration logic on
dynamic objects that cannot be iterated over in a more appropriate manner.


rexm:
Pushed NuGet 1.3.1

PR #50, opened 2015-04-28
Exception includes name of partial

danielmarbach:
+ Exception includes now name of partial
+ Use `as` instead of `is` and then `as`

I added a few changes which makes finding the problem as a user of the library a bit simpler plus a small "optimization"


danielmarbach:
Awesome! Thanks

PR #58, opened 2015-06-28
Lazy-load Handlebars environment to ensure thread safety. [re-submit with expression body style property removed]

ballance:


rexm:
This is now included in NuGet 1.3.5

PR #62, opened 2015-07-20
Add support for backslash escaping templates from evaluation

JMontagu:
Moustache/Handlebars allows a backslash to escape template evaluation
````
\{{myValue}}
````

This PR adds that ability also.


JMontagu:
Looks like the build server has failed due to a missing NuGet package, causing the test project to not build.

rexm:
Looks like nuget just puked, that package is a valid reference. Re-ran and all is good. Thanks!

rexm:
Updated NuGet to 1.4.0

PR #63, opened 2015-07-24
More IDictionary<,> support and IReadOnlyDictionary<,> support

leniency:
I added I more tests and support for Dictionaries with key types other than string. I also added in support for IReadOnlyDictionary, which requires .Net 4.5+. To handle this, I created a secondary Handlebars.Net40 proj that targets that and PCL. The nuspec file is setup to include both into framework folders.

If splitting out into multiple csproj files to handle the different frameworks seems too much, I can drop support for IReadOnlyDictionary<,> and just keep the expanded IDictionary<,> stuff.

All unit tests passing locally.

### DIFF
--- a/Handlebars.Net.nuspec
+++ b/Handlebars.Net.nuspec
@@ -14,6 +14,7 @@
         <releaseNotes>Note the root namespace changed from Handlebars to HandlebarsDotNet</releaseNotes>
     </metadata>
     <files>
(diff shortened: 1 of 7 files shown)


leniency:
Looks like some of the whitespace tests are failing in CI, but working locally. Will investigate.

rexm:
Yes, still investigating that myself as well

rexm:
The build has been fixed; if you reconcile with the master it should be good to merge.

PR #65, opened 2015-07-29
Fix numeric args getting stripped of first char

freakingawesome:
After the first argument to a helper, all subsequent numeric literal arguments were being stripped of their first character. For example, `4567` was passed to the helper as `567`.


freakingawesome:
The Travis build failure seems unrelated to my change. Looks like it's been failing with that error prior to my commit.

rexm:
Yes that's a known issue we are working on that cropped up a few days ago. Thanks :)

rexm:
The build has been fixed; if you reconcile with the master it should be good to merge.

freakingawesome:
I rebased on `master` but the Travis build still seems unhappy with the project as a whole. The error looks unrelated to my pull request.

rexm:
Looks like Travis just puked. Re-ran and looks good.

PR #68, opened 2015-08-03
Build def update for #63 from @leniency

rexm:


rexm:
I've pushed a pre-release v1.5.0-beta to NuGet. @leniency @dejx - this has impacts to how we handle PCL, can you please test that version and give feedback here?

rexm:
OK, I backed out this changeset. Feel free to re-submit when you're ready @leniency. Thanks @dejx!

dejx:
Was this properly merged?
i'm looking at the nuspec file https://github.com/rexm/Handlebars.Net/blob/master/Handlebars.Net.nuspec

There is only one file being added into nuget package, while diff on this had two, this is related to https://github.com/rexm/Handlebars.Net/issues/78 and it would solve this issue

rexm:
I'll check this weekend.

PR #69, opened 2015-08-09
Better Dictionary<,> support

leniency:
Added in expanded support for Dictionary<,> and non-string keys. Also plays nice with and without square brackets. A whole bunch of unit tests for this too.


rexm:
NuGet 1.4.4

PR #75, opened 2015-08-25
Added MVC ViewEngine / Layout support

mcintyre321:
Hi,

I'm pretty happy with this as a view engine. I noticed that the build failed, I'm imagining due to ASP.NET not being supported. 

I could separate the ASP.NET projects into another repository (which I could maintain if you prefer), leaving just the changes in core necessary for layouts and stuff?


rexm:
Cool... thanks for submitting this. We should make a few adjustments probably, in line with what you suggested. I'll ponder it some more today and get back with my thoughts.

mcintyre321:
Would you like me to break out the ASP.NET stuff into another repo? It's really the core changes I made (parser, abstracted view engine stuff for locating other views) I would like on Nuget in order to get on with things.

rexm:
Yeah, it would be easier on the maintenance of the core to separate into another repo & extension library (Handlebars.Net.Mvc or something...)

Issue #74, opened 2015-08-21
Reflection performance enhancements


rexm:
Overall, template rendering time is good - usually in the 10-15ms range for templates and data models of average size and complexity. But we can improve...

Profiling shows almost all execution time is spent in the [PathBinder, more specifically the AccessMember routine](https://github.com/rexm/Handlebars.Net/blob/master/source/Handlebars/Compiler/Translation/Expression/PathBinder.cs#L138).

Should evaluate ways to improve performance in this routine.

gretro:
Hey.

I started to work on it. I actually need the performance boost for a project I'm working on. Combined to a ViewEngine for MVC, a page I need rendered takes 2 sec, which is way too long.

I started implementing the Generic dictionary part with ExpressionTrees. I was thinking of building a cache of certain types of IDictionary<,> that are most used. I was thinking of IDictionary<string, string> and IDictionary<string, object>. Would you have other types you think are used really often?

If you have any other suggestion, I will take happily take them.

rexm:
2 seconds is quite a long time. Have you profiled to show the most delay is indeed in this routine.

gretro:
To be honest, I don't think I've dwelved that far in the handlebar dll. I just wanted confirmation at that point that it was Handlebars .Net. I decided to fix it on my own. The page was rendered using a Dictionary&lt;string, object&gt; containing a sub-dictionary containing a least 100 elements.

What leads me to think it's this method is that I saw an improvement on performance when I used a {{#with }} block inside the {{#each }} block. After I'm done refactoring, I will make sure I run some performance tests on especially big generic dictionaries and big objects.

gretro:
Oh, and I started a Unit test project for the modifications I'm making. I hooked NUnit, FluentAssertion and Faker.

rexm:
There is already a unit test project, why would we not want to add to it?

gretro:
The existing test project is composed of integration tests which test the system as a whole. They also provide very good use cases for anyone who want to start using the library. However, the other test project is really just about unit tests. It needs to be marked "friend assembly" since most classes are internal.

It seems just easier to split them, but I don't feel strongly about this. If you really prefer only one test project, I can merge them in my fork, so it will already be done when I do a pull request.

rexm:
The stats in #82 don't seem like a worthwhile net improvement. Performance improvements should start with profiling test scenarios.

PR #81, opened 2015-11-06
Reflection performance enhancements #74 refactor

gretro:
Hello rexm.

I know it's been a while since my last message in the issue itself, but I've implemented a solution using Expressions.

However, I'm not sure if the performance improvement will be good as expected. I don't really have the time to test that right now since I have a gazillion projects at once, but maybe could you test for the performance improvement?

Thanks.


.travis.yml
```diff
@@ -6,6 +6,7 @@ install:
 script:
   - xbuild /p:Configuration=Release source/Handlebars.sln
   - mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe ./source/Handlebars.Test/bin/Release/Handlebars.Test.dll
+  - mono ./testrunner/NUnit.Runners.2.6.4/tools/nunit-console.exe ./source/Handlebars.UnitTests/bin/Release/Handlebars.UnitTests.dll
 notifications:
   email:
-    - rex@rexmorgan.net
+    - gabriel.lemire@gmail.com
```
gretro: Please disregard this. I forgot to remove it for the merge...

gretro:
I also wanted to discuss the "caching" strategy to use here. We want to avoid having to compile an expression more than once because it is expensive. 

Ideally, I think it should be tied to the template itself, but the constructed expressions are now stored in static collections. You can see that in the implementations of `IMemberAccessor`. What do you think we shoudl do? Is that good enough?

PR #82, opened 2015-11-06
Reflection perf

gretro:
Hello rexm.

I know it's been a while since my last post on the issue #74, but here is a version of the improvement. Do you have an actual test suite for measure heavy usages? It would be great if you could do a before / after measurement.

I also wanted to point out an issue I'm not sure how you want to resolve. If you take a look at the implementations of the new interface `IMemberAccessor`, you will see that many of them actually have a static Dictionnary to hold a reference to a compiled accessor to avoid having to recompile it which is costy.

Ideally, this *caching* should maybe be associated with the template itself. However, because I didn't really have the time to implement it, I just went for the easy solution for now.

I await your review and your opinion on the matter.
Thanks.


rexm:
This will take me a little time to peer review & perf test.

gretro:
Hey rexm.

Do you have any preliminary comments on the code? I think that dictionnaries used to cache the compiled expressions should be `ConcurrentDictionnary`. However, before doing this modification, I'd really like to know if it makes a big difference in the time of execution.

rexm:
Still doing some evaluations on different perf conditions.

ejsmith:
Have you had a chance to look through this? I think this is a really nice improvement and makes it pluggable for adding support for more object types like JObject.

gretro:
There is one thing that REALLY bugs me about this solution: The framework is no longer thread-safe. Because we target the lower common denominator, we don't have tools like `ConcurrentDictionnary`. The odds of being multi-threaded on a mobile or on Silverlight is lower, but it's not non-existant. I would really like to avoid having to implement something using locks. I pretty much suck at multi-threading safety.

Does anyone have any idea?

ejsmith:
Could use something like this:

http://stackoverflow.com/a/1336582/55890

gretro:
@ejsmith I like the solution. I didn't even know this existed. However, this is a remaining problem. Because the dictionnaries are static, they will "leak". If you have a scenario where you use a template only once or twice, but you have a lot of them, you will end up leaking.

The solution I have found is to cache the accessors inside the `HandlebarsConfiguration.cs` as an internal property. This could solve the problem since it could allow users to create different configurations when they have a similar scenario, or use the default one for templates that are used more often. What do you think? I'm I going too far or could this be a real need?

ejsmith:
I wouldn't think that would be a real world problem, because the number of objects that are being passed into templates I would think is a finite number of items for each application. Anything that is more dynamic is something like what I am doing with using JObject which wouldn't cache the accessors in a dictionary.

gretro:
If you need to inject the `JObject` accessor as a plugin, I could probably put it in the `HandlebarsConfiguration.cs` as a public property. This way, you could create your own accessor without worrying about creating a dependancy between Handlebars.Net and Json.Net.

What do you think @rexm? Could this fit in the Handlebars.Net vision?

ejsmith:
Yeah, should definitely add a configuration property for the member accessors.

ejsmith:
@gretro are you going to work on these changes? Would be easier for you to make them, but I can work on it if you want.

gretro:
I actually meant to put some hours on it. I started a merge back a few weeks back, but the tests were broken. Since then, I haven't put any time on it. The changes we spoke about could be implemented quite easily actually.

I would work on it at evenings. When do you need this feature for?

ejsmith:
Doh. Looks like we are both working on it.

https://github.com/learningmachine/Handlebars.Net/tree/ReflectionPerf

ejsmith:
Don't think I like the NullInstanceMemberAccessor. Doesn't seem like the member accessors should have to be responsible for filtering out null values.

I am also thinking that we should change up the interface to pass in a ResolveMemberContext object that gives the member accessor all of the information it will need to have access to instead of passing only the instance and member name parameters.

gretro:
The solution provided on Stack Overflow (http://stackoverflow.com/a/1336582/55890) doesn't work since we also target Silverlight 5, which has no support for this class.

ejsmith:
Could we remove Silverlight 5 since that isn't really a platform that I would think that a Handlebars lib would be very useful and also since Silverlight is pretty much dead?

gretro:
@ejsmith I have no strong feeling regarding the creation of a `ResolveMemberContext` object. The cache could be passed in this object. This would get rid of the `RequiresResolvedMemberName` property, which is ambiguous.

As for the `NullInstanceMemberAccessor`, this could be the responsibility of the `PathBinder` and the `IteratorBinder` to check for null. Traversing all accessors to get a simple `UndefinedBindingResult` is kind of a pain to my sense. This could be done at the top of the methods.

ejsmith:
Yeah, I think we could give access to the instance, member name, configuration, and whatever else we decide that we will need. Will be able to add new things without breaking the member accessor implementations in the future as well.

gretro:
@ejsmith Sold!

ejsmith:
:+1:

rexm:
Two questions: 

1. Have you established the degree of performance improvement this yields? Range of factor x.x to y.y for some set of scenarios?
2. If a feature implementation has a potential for silent leakage (i.e. non-obvious that it's accumulating objects over time), should it be opt-in and how would you accomplish that?

gretro:
Hey rexm.

1. No, I haven't done that yet. However, this refactoring is probably still better than the big fat method that was used to access the member before. Do you have certain performance tests or an approach you use for performance testing?

2. There is a potential for "silent" leakage because we'll store the compiled expression methods to avoid recompiling them. However, these methods should be stored in the `HandlebarsConfiguration`. Maybe should we create a method to flush this cache. However, let's say you compile a template and you execute it over and over again with the same types, the memory footprint will remain the same. Opting-in or out is not really an option. If you were to opt-out, the performance would just hit rock bottom. Also, we have to note that only a few scenarios require this cache: Generic dictionnaries and Object members.

rexm:
I'd start by using the exact scenario that led you to conclude the current implementation is too slow.

BenSwayne:
> I'd start by using the exact scenario that led you to conclude the current implementation is too slow.

Seconded! Start with a reproduction case of the problem - then architect the solution. Not the other way around IMHO.

gretro:
I created a template which iterates over a collection, and I augment the size of this collection. The object in the collection has HashMaps, object reflection access, Generic dictionnaries, collections and dynamic objects (Expando).

The refactor is slightly faster in most cases (except when there is only one occurence in the collection). It shows a bit with large and very large collections (1000 - 5000 instances). 

I used dotTrace to see what was slow (13 sec for 5000 instances). What is really slow is accessing dynamic objects, no matter if I use the new implementation or not.

<pre>
{
    "1": {
        "old": "0:00:00.0245518",
        "new": "0:00:00.0343023",
        "old-no-dynamic": "0:00:00,0206545",
        "new-no-dynamic": "0:00:00.0258079",
        "new-no-dynamic-no-cache": "0:00:00.0352642"
    },
    "100": {
        "old": "0:00:00.2913603",
        "new": "0:00:00.2890101",
        "old-no-dynamic": "0:00:00.0555818",
        "new-no-dynamic": "0:00:00.0470506",
        "new-no-dynamic-no-cache": "0:00:01.0846047"
    },
    "500": {
        "old": "0:00:01.4175398",
        "new": "0:00:01.3991529",
        "old-no-dynamic": "0:00:00.2405773",
        "new-no-dynamic": "0:00:00.1795344",
        "new-no-dynamic-no-cache": "0:00:04.8844109"
    },
    "1000": {
        "old": "0:00:02.8154299",
        "new": "0:00:02.8154245",
        "old-no-dynamic": "0:00:00.4547153",
        "new-no-dynamic": "0:00:00.3443646",
        "new-no-dynamic-no-cache": "0:00:09.8250836"
    },
    "5000": {
        "old": "0:00:13.9174308",
        "new": "0:00:13.7143033",
        "old-no-dynamic": "0:00:02.2296616",
        "new-no-dynamic": "0:00:01.6976068",
        "new-no-dynamic-no-cache": "0:00:49.1135779"
    }
}
</pre>

As I said, with no cache, the performance of the new implementation just really suck. The performance of going through reflection is really not bad in most cases. It really only makes a difference in really huge collections. I think we should still go with the refactor to allow plugin-style accessors (to allow reading from JObject, for instance), but I would tend to remove the solution with expressions. It complexifies the code for no good reason.

What do you think? I'm pushing my test.

ejsmith:
I would imagine that these changes make a good size difference when rendering many templates of normal sizes. In my app, I need to iterate through potentially 100s of thousands of items and render them.

gretro:
Let's continue the discussion in issue #74.

PR #90, opened 2016-01-20
Add instance to ResolveExpressionName to allow advanced scenerios

ejsmith:
I know this is a breaking change, but not having any contextual information to control the resolved name is a blocker for my use case.  Any ideas on how to implement this without it being a breaking change?

### DIFF
--- a/source/Handlebars.Test/DynamicTests.cs
+++ b/source/Handlebars.Test/DynamicTests.cs
@@ -93,20 +93,6 @@ namespace HandlebarsDotNet.Test
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


ejsmith:
Oops I didn't mean to add the broken test to this pull request. I am investigating ways to fix this test.

rexm:
I'm curious what the underlying problem case is

ejsmith:
The problem is that my normal naming convention is lower_case for json, but I also have unknown data types added to a dictionary that can be any case.

ejsmith:
Going to delete this one and resubmit correctly

PR #91, opened 2016-01-21
Add support for JObject's to #if

ejsmith:


ejsmith:
Let me know what you think about this solution. Obviously, we don't want to add a reference to JSON.NET, so I am just checking for the type name and having to use a little reflection.

rexm:
This seems more problematic than taking the binary dependency. Is this a problem you can address in the app with an adapter pattern?

ejsmith:
I definitely think the library should support using JObject as the data source, but I didn't really think it was worth it to take on the dependency directly. If you would rather take a dependency on JSON.NET then that is fine.

Not sure what you mean about the adapter pattern. Do you mean that I should try and adapt my source object into a different type that would work better with your lib?

rexm:
Yes, JSON.NET has a number of conventions for producing object graphs that are convertible to BCL types, discussed as recently as yesterday - https://github.com/rexm/Handlebars.Net/issues/88

ejsmith:
I just looked through pull #82 and I think this could be implemented as a IMemberAccessor. We would just need to make sure that end users can plug in custom member accessors through a configuration setting. Then we could just add a simple JObject member accessor that uses JSON.NET directly.

rexm:
PR #82 is closed, but that idea seems like a reasonable path forward.

PR #92, opened 2016-01-21
Add instance to ResolveExpressionName

ejsmith:
My naming conventions are lower_case for my json objects, but I have unknown items in dictionaries that I need to leave the casing alone on. Without access to the contextual information of what object I am resolving the member name for, I don't know what format to use.


rexm:
This is a net improvement and implementors should upgrade their contracts accordingly. It's technically a breaking change but a simple contract change is the fix; minimal work required.

rexm:
Updated NuGet 1.5.3

PR #93, opened 2016-01-28
Add support for passing hash parameters to partials

amaclean:
This library is awesome, but I really wanted to be able to pass a bunch of parameters to a partial that were not in the root context (http://handlebarsjs.com/partials.html#partial-parameters).

I've added a new HashParameters expression that contains a collection of key value pairs (string, object) to achieve this. It supports the value being a string, int, bool or a context path.

e.g. `{{>person pie=1 foo=true bar="hello" name=leadDev.name }}`

I didn't have a massive amount of time to spend but I have written a bunch of unit tests and everything looks good.

Let me know what you think.

### DIFF
--- a/source/Handlebars.Test/PartialTests.cs
+++ b/source/Handlebars.Test/PartialTests.cs
@@ -74,7 +74,7 @@ namespace HandlebarsDotNet.Test
         [Test]
         public void BasicPartialWithMultipleStringParameters()
         {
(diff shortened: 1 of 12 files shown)


simmo:
+1 👍 This would be an excellent addition! Nice work @amaclean

rexm:
Looks great... thanks for the PR

rexm:
Updated NuGet 1.6.0

PR #98, opened 2016-02-20
Add support for dynamic partials

amaclean:
The change to `Tokenizer.cs` was to move the sub-expression end token higher as this seemed to get skipped when arguments were added to the sub-expression.

### DIFF
--- a/source/Handlebars.Test/PartialTests.cs
+++ b/source/Handlebars.Test/PartialTests.cs
@@ -166,6 +166,87 @@ namespace HandlebarsDotNet.Test
         }
 
         [Test]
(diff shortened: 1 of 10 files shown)


rexm:
Do we have test coverage for passing hash args after the sub expression?

amaclean:
No, just regular arguments at the moment.

I've created one and it's failing unfortunately, it seems that the ')' at the end of the sub-expression is getting swallowed during

amaclean:
So it looks like hash arguments on sub-expressions didn't actually work anyway!

The fix is simple so I'll create a new PR.

PR #99, opened 2016-02-20
Fix hash arguments on sub-expressions.

amaclean:


rexm:
Will push to NuGet later tonight

PR #101, opened 2016-02-25
Update WordParser to allow for expression chars within a string param…

amaclean:
A minor change to the WordParser to allow for certain expression characters to be included in a string parameter.


rexm:
Thanks, updated Nuget to 1.6.4. Just out of curiosity, did you find this gap from an actual use case you've got?

PR #120, opened 2016-05-20
Add injectable text encoder

imsedim:
By default Handlebars HTML-escapes values returned by a {{expression}}. This patch adds the possibility to configure the custom text encoder. It is also possible to run Handlebars without encoder - just like "noEscape" otpion in Handlebars.js.


imsedim:
I felt a bit adventurous and decided to refactor the existing code to try to simplify it a litle bit. 

So I added a refactoring commit - it does not add any functionality (so feel free to ditch it if you're unhappy with changes).

rexm:
Released in nuget version 1.7.0

rexm:
@imsedim Thanks for the PR, and nice job on the refactor - much cleaner.

PR #124, opened 2016-05-25
support else in #section

imsedim:
It is now possible to use {{else}} within {{#section}}, similar to mustache and handlebars.js:

    {{#obj}}
      {{objProperty}}
    {{else}}
      inversion
    {{/obj}}
and

    {{^obj}}
      nothing
    {{else}}
      {{objPropery}}
    {{/obj}}


imsedim:
Added fix of bug #125 to this PR (since I already rewrote the affected methods here)

rexm:
Looks good, thanks! Will get this into a new version of the nuget package tonight.

PR #128, opened 2016-05-27
Fix whitespace bug with dictionary support

frxnz:
Fixes some whitespace issues with dictionary support. The following now work:

```handlebars
{{ [expression] }}
{{[expression]}}

{{ parent.[expression] }}
{{parent.[other expression]}}
```

### DIFF
--- a/source/Handlebars/Compiler/Lexer/Parsers/WordParser.cs
+++ b/source/Handlebars/Compiler/Lexer/Parsers/WordParser.cs
@@ -7,7 +7,7 @@ namespace HandlebarsDotNet.Compiler.Lexer
 {
     internal class WordParser : Parser
     {
-        private const string validWordStartCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_$.@";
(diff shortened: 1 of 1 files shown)


source/Handlebars/Compiler/Lexer/Parsers/WordParser.cs
```diff
@@ -45,7 +45,7 @@ private string AccumulateWord(TextReader reader)
                 {
                     var peek = (char)reader.Peek();
 
-                    if (peek == '}' || peek == '~' || peek == ')' || char.IsWhiteSpace(peek))
+                    if (peek == '}' || peek == '~' || peek == ')' || (char.IsWhiteSpace(peek) && !buffer.ToString().Contains("[")))
```
rexm: I don't think this is a safe assumption. We shouldn't merely check to see if the buffer contains a `[`, but we should check to see if the buffer contains a `[` but has no subsequent `]` to ensure we are inside the brackets.

frxnz:
@rexm Added a check for the closing square bracket

muizsyed:
+1 Any idea when this will be merged?

rexm:
Pushed to nuget 1.7.1

PR #134, opened 2016-06-10
use _configuration for locks instead of Handlebars type

pgrm:
As mentioned in https://github.com/rexm/Handlebars.Net/issues/131 I went for reusing _configuration for locking instead of creating a separate lock variable, because _configuration is used within those locks and probably two instances with the same underlying configuration shouldn't interfere between each other.


rexm:

Issue #141, opened 2016-07-05
Partials ignore context if used with string parameters


DmitryLobanov:
Imagine such test added to the `PartialTests` class:

```
        [Test]
        public void BasicPartialWithStringParametersAndImplicitContext()
        {
            string source = "Hello, {{>person last='Smith'}}!";

            var template = Handlebars.Compile(source);

            var data = new
            {
                name = "Marc"
            };

            var partialSource = "{{name}} {{last}}";
            using (var reader = new StringReader(partialSource))
            {
                var partialTemplate = Handlebars.Compile(reader);
                Handlebars.RegisterTemplate("person", partialTemplate);
            }

            var result = template(data);
            Assert.AreEqual("Hello, Marc Smith!", result);
        }
```

This test fails, because partial fails to get parameter `name` from context. But, according to the example on the [Handlebars](http://handlebarsjs.com/) page (section "Partials") this should work. So it seems that if we pass string parameters to partial template, it "forgets" about the context.

Oh, also, if we have partial called from that partial, they both don't get the context. But if we remove string parameters, then context gets passed both to partial and even into partial, included into the first partial.

rexm:
Thanks for the bug report. I don't have a timeframe for a fix yet; but PR always welcome.

tsliang:
**tl;dr Partials don't access their parent BindingContext, and it can be fixed in PathBinder (less effort, possible side effects), PartialBinder (much more effort, but with no side effects), or PartialConverter (medium effort, probably no side effects)**

I've been investigating this issue (as it affects my project as well), and I've come up with the root of the issue, along with possible solutions. 

The problem starts in PartialBinder.VisitPartialExpression:

```csharp
protected override Expression VisitPartialExpression(PartialExpression pex)
        {
            Expression bindingContext = CompilationContext.BindingContext;
            if (pex.Argument != null) // In this test, this is a dict of {"last": "Smith"}
            {
               // Binding context becomes {"last": "Smith"}, with the root BindingContext as a parent
                bindingContext = Expression.Call(
                    bindingContext,
                    typeof(BindingContext).GetMethod("CreateChildContext"),
                    pex.Argument); 
            }
            return Expression.Call(
                new Action<string, BindingContext, HandlebarsConfiguration>(InvokePartial).Method,
                Expression.Convert(pex.PartialName, typeof(string)),
                bindingContext,
                Expression.Constant(CompilationContext.Configuration));
        }
```
[...]

PR #149, opened 2016-08-28
Allow partials with hash parameters to fallback to parent context

amaclean:
This is a potential fix for #141 for you to look at.

It's simple but not amazingly elegant. It involves a new class to represent a hash parameter dictionary to prevent any false positives, then allowing `PathBinder.ResolvePath` to fallback to the parent context if hash parameters are used as the current context.

I have also added a test to make sure that when a child context with an incomplete scope is used it does not fallback to the parent context, to make sure we still match current handlebars behaviour.


rexm:

PR #150, opened 2016-08-29
Fall back to parent context if value not found in partial parameters

tsliang:
In cases when a partial is called with literal arguments, but a variable within the partial is not part of those literal arguments, the partial should attempt to resolve the value from the parent context.

Fixes Issue #141.

### DIFF
--- a/source/Handlebars.Test/PartialTests.cs
+++ b/source/Handlebars.Test/PartialTests.cs
@@ -54,7 +54,35 @@ namespace HandlebarsDotNet.Test
             Assert.AreEqual("Hello, Marc!", result);
         }
 
-        [Test]
+		[Test]
+		public void BasicPartialWithIncompleteContextIgnoringParentContext()
+		{
+			string source = "Hello, {{>person leadDev}}!";
+
+			var template = Handlebars.Compile(source);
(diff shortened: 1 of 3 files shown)


rexm:
This seems potentially problematic on the face of it, because it kind of reimplements a simple/limited version of what the PathBinder does.

tsliang:
Does it? I don't see where. The PathBinder helps to resolve some of the paths that might be contained in the HashParametersExpression, and these paths come into MergeContextValues already resolved. No actual functionality is duplicated - MergeContextValues just combines the already-resolved Context value and hash params. 

Maybe the logic could be refactored. It does seem a little too general for something as specific as the PartialBinder, but then again, this is a partial-only problem...

rexm:
Yes... for example, what if the original model is dynamic, or has fields instead of properties? These kinds of cases are handled by the PathBinder but those cases fail or would be lost during the merge.

tsliang:
Ahh right, of course. So in that case, we could move the merge code to the PathBinder, which seems like an overreach, but could also save us some trouble. What do you think?

rexm:
This was addressed in #149

PR #152, opened 2016-09-08
Fix for empty partial parameter incorrectly falling back to parent

amaclean:
In JS a partial will not fallback if a parameter has been passed and is empty (http://fiddle.jshell.net/jatbc1mu/5/).

This was not the case with my initial change to add the fallback behaviour  (#149).


rexm:

PR #161, opened 2016-10-24
Exposing Internal Compiler and Environment

mikeofireland:
I needed the ability to expand on the base compiler to give a more tuned ViewEngine object for MVC .net.  I am testing my code and did not want to rewrite your compile engine.  Basically, this is a start on the ability to build custom compile engines.  I think the CompileView looks like a pretty custom compile engine and think it should be moved into an extension of the base compile class.  Anyway, that is a tangled web and this at least exposes the first layer of the compile engine to expand off of.

The changes will:

- Expose HandlebarsEnvironment to the public.  Inheritable and possible build your own with IHandlebars.
- Expose Token to the public through IToken interface.
- Expose HandlebarsCompiler to the public through IHandlebarsCompiler interface.
- Expose 3 new functions in HandlebarsCompiler that make up the primary method of compiling.

The three new function in HandlebarsCompiler:

1. IEnumerable<IToken> Tokenize( TextReader source )
2. IEnumerable<Expression> ExpressionBuilder( IEnumerable<IToken> tokens )
3. Action<TextWriter, object> FunctionBuilder( IEnumerable<Expression> expressions, string templatePath = null )

### DIFF
--- a/source/Handlebars/Compiler/HandlebarsCompiler.cs
+++ b/source/Handlebars/Compiler/HandlebarsCompiler.cs
@@ -10,7 +10,7 @@ using HandlebarsDotNet.Compiler.Lexer;
 
 namespace HandlebarsDotNet.Compiler
 {
-    internal class HandlebarsCompiler
+    internal class HandlebarsCompiler : IHandlebarsCompiler
     {
         private Tokenizer _tokenizer;
         private FunctionBuilder _functionBuilder;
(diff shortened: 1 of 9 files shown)


rexm:
I'd like to understand the roadmap for this approach... for example how does abstracting the compiler itself and some of its internal constructs move towards an ASP.NET MVC viewengine imementation?

mikeofireland:
I am expanding on the base implementation of file system and the layout system with more customization's on it.  Particularly how layouts are compiled.  Basically, I rewrote "compileView" with my own implementation of it.

CompileView is very static in the current implementation and I am looking to customize it quite a bit more.  It looks like to me CompileView should be an extension off the base compiler instead of its own implementation based deep in the engine.  I don't want to rewrite the compile class and pull the compileView out of it so I took the easiest path of exposing the compile engine.

I also think the filesystem should be expanded on so we have more customization on the deeper engine features (templatePath for compile and how it auto finds partials based on filesystem).

The problem at the core:  HandlebarsCompiler is not exposed to a custom HandlebarsEnvironment.  I would rather not fork then start rewriting the whole compiler portion of the repo.  It would be easier to just expose parts and allow customization on top of the base compiler.

My project:  Currently, I need a HandlebarsCompiler that can compile views in .net MVC.  The idea would be to provide a different project that uses Handlebars.Net as a base and provides a view engine on top of the base functionality.  So you could see a Handlebars.Net.MvcViewEngine and a Handlebars.Net.NancyViewEngine.

I work with both Nancy and MVC . net and would want to build out a view engine for both.

mikeofireland:
> how does abstracting the compiler itself and some of its internal constructs move towards an ASP.NET MVC viewengine implementation?

The current implementation is very tough to work with in terms of file system and how "layouts" work.  Its not built out all the way and I want to provide my own implementation of it.  With my own implementation I want to use some of the features that are built into the engine.  Like how file system configuration works but with further advancements on it.

This is what I got working based on my fork:
- Can call HbsView from controllers and return a MVC.ActionResult
- Customized filesystem to use MVC like paths customization ( Example: see ViewLocationFormats )
- Reworked how the view, model, and layout interact so that its easier to see what data renders where.

rexm:
Where are these implementations you describe?

mikeofireland:
This might be easier to show not try to explain.  I will be working on getting my implementation up and going.

PR #169, opened 2016-11-21
Keep track of block/helper name in exceptions

JvanderStad:
If an exception occurs at runtime, show helper/blockname in exception


rexm:
Looks like it needs a rebase

JvanderStad:
Rebased

PR #170, opened 2016-11-21
Pathbinder with null object caused exception

JvanderStad:
Pathbinder with null object caused exception


JvanderStad:
Also did some code cleanup

Issue #107, opened 2016-03-23
Getting the @index / @first / @last from a root context


breenyoung:
Maybe I'm gettin the syntax wrong, but anyone know if when you're in a  nested each loop you can get at the parent's context of @index/first/last in the parent loop?

`{{@each}}`

`Parent index: {{@index}}`

```
{{#each}}

    How to get at parent @index?
    {{../@index}} ?
    {{@root.index}} ?

{{/each}}
```

`{{/each}}`

rexm:
Need to figure out why, `{{../../@index}}` will get you the current parent index. That isn't correct or intuitive, but that's how it seems to work right now.

rexm:
How does handlebarsjs work in this situation?

imsedim:
From [here](http://handlebarsjs.com/builtin_helpers.html)

> Nested each blocks may access the interation variables via depth based paths. To access the parent index, for example, {{@../index}} can be used.

304NotModified:
Isn't this fixed with #174?

PR #174, opened 2016-12-05
Fix issue with getting a variable from a parent context

abraham-fox:
Nested loop now can get variables from parent context as it works in handlebars.js (Issue #107)


rexm:
Thanks for the PR... implementation looks good. Can you replace the tabs with 4 spaces before we merge?

abraham-fox:
Sure I can! Done)

PR #184, opened 2017-01-30
Iteration improvement

Andrew-Hanlon:
Changed the iterate method to not cause double-enumeration of true IEnumerables in order to get the Count/Length for knowing the last item - or in the words of the Todo note above the method "make this a little less dumb".

The foreach is unrolled to a while loop that 'caches' one item ahead to know when it has reached the last item without knowing the overall length.

All tests pass.


rexm:
Good change. Thanks!

Andrew-Hanlon:
Happy to contribute - and many thanks to you for this great library!

PR #206, opened 2017-04-05
Update README.md

Tombert:
I think the markdown will render prettier if you add spaces to the header blocks.


rexm:
This is duplicating the effort from #205

Tombert:
Ah, didn't see that.  I'll close the request then.

PR #213, opened 2017-05-08
Set TFM to minimum of net452

jchannon:
The minimum full fwk now supported is net452


rexm:
What was the driver for this change?

jchannon:
net452 is the oldest fwk version that is now supported

jchannon:
https://blogs.msdn.microsoft.com/dotnet/2015/12/09/support-ending-for-the-net-framework-4-4-5-and-4-5-1/

rexm:
Ah, forgot about that. Thanks

PR #220, opened 2017-07-06
Don't throw undefined binding exception when value is null

hmflash:
Commit 53e8b5315cbe5b1ed43123775509988de1e8f642 adds `ThrowOnUnresolvedBindingExpression` which I want to use to validate my templates.  However, the current implementation will throw a `HandlebarsUndefinedBindingException` when the property exists but is null.  This PR fixes that.

I am trying to use `HandlebarsUndefinedBindingException=true` with the following template:
```
{{#if Data}}
Operation: {{Data.Operation}}
{{/if}}
```


rexm:
Thanks!

PR #221, opened 2017-08-01
Added support for inline partials. People that are more familiar with…

mikeprince3:
… the codbase should look at some of the hackier bits. I left comments to explain the thinking in the code.

The template below can be used as a simple test. I have done some pretty complicated nesting of inline partials with parameters and different contexts and everything seems to work pretty well. I'm sure some aspects of this can and should be improved so please feel free to put things where they should be.

{{#*inline "inlineTest"}}
    <strong>hello inline partial</strong>
{{/inline}}
{{> inlineTest}}

### DIFF
--- a/source/Handlebars.Test/InlinePartialTests.cs
+++ b/source/Handlebars.Test/InlinePartialTests.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using Xunit;
-using System.IO;
+﻿using Xunit;
 
 namespace HandlebarsDotNet.Test
 {
--- a/source/Handlebars.Test/InlinePartialTests.cs
+++ b/source/Handlebars.Test/InlinePartialTests.cs
@@ -1,0 +1,343 @@
+﻿using System;
+using System.Collections.Generic;
+using Xunit;
+using System.IO;
+
+namespace HandlebarsDotNet.Test
+{
+    public class InlinePartialTests
+    {
+        [Fact]
+        public void BasicInlinePartial()
+        {
+            string source = "{{#*inline \"person\"}}{{name}}{{/inline}}Hello, {{>person}}!";
+
+            var template = Handlebars.Compile(source);
+
+            var data = new
+            {
+                name = "Marc"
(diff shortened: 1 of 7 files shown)


rexm:
This is a pretty effective solution. Only 1 aspect gives me discomfort... the empty built-in helper. What problem does that solve? Otherwise, it seems fairly straightforward. In terms of the base syntax, an inline partial more or less _is_ a block helper, so this approach makes sense. We do need some unit test coverage to call it complete, however.

mikeprince3:
I don't like the empty helper either but it seemed to help side-step understanding a big chunk of code. I also suspect the addition to the BlockHelperFunctionBinder.VisitBlockHelperExpression is a bit of a hack too. I'll admit to being new Expressions too so this was a simple way for me to get it working without hitting that learning curve.

Given your more thorough knowledge of the code base, if you can provide some guidance as to how to accomplish the same things, I am happy to take another shot at it and add some tests as well.

yringler:
+1. This is a super awesome and super useful; any template of size is much more maintainable when broken to components, and it's nice not to have to clutter the file system with a handful of partials and the c# code with all the registering.
I bumped into this thread when I realized that this wasn't already supported.
Hope this is released to NuGet soon.

rexm:
I think if you can add some unit tests, this is good enough to go as-is. It works fairly well and after thinking about the approach more, I dig the simplicity because inline partials are syntactically a weird kind of block helper.

mikeprince3:
While testing, it turned out that using Handlebars.RegisterTemplate would pollute the static space and could collide with other templates that were compiled with the same instance.  This was only an issue with Handlebars.Compile. It worked fine with Handlebars.Create. That said, it seemed more appropriate that the inline partials should be scoped to the compiled template and nothing else. To do this, I added a collection of compiled inline partials to pass around with the BindingContext. 

I duplicated the existing Partial test into InlinePartial tests and everything is passing.

mikeprince3:
On a side note... thanks for putting this project together. It is solving some problems I have in a very good way. Great code.

PR #232, opened 2017-10-25
Migrate test to HtmlAgilityPack (more maintained)

jokin:
CsQuery hasn't been updated in the past 2 years. HtmlAgilityPack it's currently mantained. HtmlAgilityPack also supports dotnetcore, while for the test project I don't know if it's relevant.


rexm:
Going to avoid any additional investment in the baked-in view engine and focus on a better extensibility model in [v2](https://github.com/rexm/Handlebars.Net/issues/294) that will afford well-designed view engines.

PR #236, opened 2017-12-04
Added Netstandard2.0

uffebjorklund:
I am on a Mac so I only compiled for netstandard, did not change any code so net452 etc should be fine. Please note that I changed version (in Handlebars.csproj) and added some lines in travis.yml and appveyor.yml that you might wanna discard.

Regards, Uffe

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,12 @@ script:
   - dotnet restore source
   - dotnet build source/Handlebars/Handlebars.csproj --configuration Debug --framework netstandard1.3
   - dotnet build source/Handlebars/Handlebars.csproj --configuration Release --framework netstandard1.3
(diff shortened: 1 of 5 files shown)


uffebjorklund:
My guess is that the build failed since it is using net sdk 1.0.1 and not 2.0.
Found a fix here: https://codeblog.jonskeet.uk/2017/08/15/using-net-core-2-0-sdk-on-travis/ by J Skeet since you probably want net core 1.0 support as well.

rexm:
Ok, waiting for green

uffebjorklund:
@rexm is it ok? If it is all good, can you release a new version to nuget after accepting?

rexm:
Will aim to get it out this weekend.

CrosserTechnologies:
@rexm do you have a estimate for releasing a new version with support for netstandard2.0 to nuget?

PR #237, opened 2017-12-07
Compile partial blocks to allow nested partial-block usage

huysentruitw:
Also took the occasion to simplify the PartialBinder logic a bit.

As a plus, this change compiles the template fallback value.


pmccloghrylaing:
👍 
Ran into this issue with nested partials - your fork works as expected for me and the tests cover nesting a combination of content and partials.
@rexm any chance this can be merged in?

chadly:
> Will release this weekend

I've just run into this issue as well, compiled the `master` branch locally and confirmed that this fixes our issue with nested partials.

Any chance for a release of the current state of `master`?

Thanks for the great library by the way, @rexm!

PR #244, opened 2018-02-23
Add IsUndefinedBindingResult

el2iot2:
Add a public util method that checks for the internal type `UndefinedBindingResult` so a registered helper can check for an undefined binding result.

### DIFF
--- a/source/Handlebars/HandlebarsUtils.cs
+++ b/source/Handlebars/HandlebarsUtils.cs
@@ -11,6 +11,11 @@ namespace HandlebarsDotNet
         {
             return !IsFalsy(value);
         }
+        
+        public static bool IsUndefinedBindingResult(object value)
+        {
+            return value is UndefinedBindingResult;
+        }
 
         public static bool IsFalsy(object value)
         {


el2iot2:
Essentially I want to write a "coalesce" helper to handle default values. Something like:
```
            Handlebars.RegisterHelper(nameof(coalesce), coalesce);
            //local function :)
            void coalesce(TextWriter output, dynamic context, object[] arguments)
            {
                foreach (var argument in arguments)
                {
                    switch (argument)
                    {
                        case string s when s != null:
                        case bool b:
                        case sbyte _:
                        case byte _:
                        case short _:
                        case ushort _:
                        case int _:
                        case uint _:
                        case long _:
                        case ulong _:
                        case float _:
                        case double _:
                        case decimal _:
                            output.Write(argument);
                            break;
                        
                        //After checking above, this is essentially
                        //if (value is UndefinedBindingResult)
                        //Which we cannot do because this is an internal type
                        case object o when HandlebarsUtils.IsFalsy(o):
                            continue;
                    }
                }
            }
```

A `HandlebarsUtils.IsUndefinedBindingResult` would simplify it quite a bit. Thanks for your consideration!

el2iot2:
@rexm a unit test seems like overkill, but I am happy to adjust as needed on that and/or naming!

rexm:
Clever. Will release an updated version to nuget in the next few days.

Issue #136, opened 2016-06-17
Version 1.7.1 of nuget package introduces breaking change


czuroski:
I am running into a template compilation issue after moving to version 1.7.1 of the nuget package.  The error is - 

"Non-empty collection required\r\nParameter name: expressions"

Unfortunately, it is difficult to get any more detail with that.  If this is not a known issue, then I can pull down the latest source next week and attempt to pinpoint the cause of the issue.
Until then, I can roll back to 1.7.0 and everything works fine.

rexm:
Can you share the template and example data?

rexm:
Closing this due to inactivity; can reopen any time.

UniSnake:
I think I reproduced this and the cause was empty helper or conditional blocks:
Fine on older Handlerbars.Net versions, throws the aforementioned exception on newer versions:
{{#myhelper}}{{/myhelper}}
{{#if}}{{else}}do this{{/if}}
{{#ifmycondition}}{{else}}do this{{/if}}

I'm not sure if empty blocks are allowed by Handlerbars.js, but I found that little bit of fault tolerance nice, and the exception message doesn't make it easy to track down what is wrong.

Issue #114, opened 2016-05-12
Space required between block helper close and open tags


206mph:
We found that this is causing the template to fail:

{{#is ProgramID ""}}no program{{/is}}{{#is ProgramID "1081"}}some program text{{/is}}

However, when we format it with a space between the block close tag and block open tag, it works.

{{#is ProgramID ""}}no program{{/is}} {{#is ProgramID "1081"}}some program text{{/is}}

Any ideas?

rexm:
I tried to reproduce this as best I can (I am not sure what the `#is` helper does, though I can guess) - and it seems to work fine. Can you provide more details? i.e. What does "fail" mean?

206mph:
The helper code:
`			Handlebars.RegisterHelper("is", (writer, options, context, arguments) =>
			{
				try
				{
					var current = arguments[0].NullSafeToString();
					var compare = arguments[1].NullSafeToString();
					if (current == compare)
					{
						options.Template(writer, (object)context);
					}
					else
					{
						options.Inverse(writer, (object)context);
					}
				}
				catch
				{
					//do something
				}
			});
`

When I call Handlebars.compile, it generates the error:

> Non-empty collection required
> Parameter name: expressions

rexm:
Only two things I can think of that have made my test cases pass and someone else's fail:
- Can you confirm that is the exact template you're testing?
- Are you using Windows (I'm on Mac, every once in awhile they are slightly different)

206mph:
This is odd.  I did a little more digging myself and can confirm that it works fine if my template has only the 2 "is" helpers with no space.  I can't provide the whole template as my client might not be happy with me.  I'll have to do more investigating to see what's causing the issue.  Meanwhile, at least we have a workaround of adding the space between the helper close/open tags.

PR #250, opened 2018-05-16
Test and fix for empty iterator block bodies

nblumhardt:
Hi! Thanks for the great library.

I've reproduced #114 and and #136, which I believe to be the same issue (both reports include the same exception message and similar conditions).

```csharp
[Fact]
public void EmptyElementTemplate()
{
    var source = "Hello,{{#each people}}{{/each}}";
    var data = new
    {
        people = new[]{
            new {
                name = "Erik"
            },
            new {
                name = "Helen"
            }
        }
    };
    var result = template(data);
    Assert.Equal("Hello,", result);
}
```

The test case shows an `#each` block with no content. The exception generated is:

```
System.ArgumentException : Non-empty collection required
Parameter name: expressions
   at System.Dynamic.Utils.ContractUtils.RequiresNotEmpty[T](ICollection`1 collection, String paramName)
   at System.Linq.Expressions.Expression.Block(IEnumerable`1 variables, IEnumerable`1 expressions)
   at HandlebarsDotNet.Compiler.IteratorBlockAccumulatorContext.IsClosingElement(Expression item)
   at HandlebarsDotNet.Compiler.BlockAccumulator.AccumulateBlock(IEnumerator`1 enumerator, BlockAccumulatorContext context)
   at HandlebarsDotNet.Compiler.BlockAccumulator.<ConvertTokens>d__3.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at HandlebarsDotNet.Compiler.BlockAccumulator.Accumulate(IEnumerable`1 tokens, HandlebarsConfiguration configuration)
   at HandlebarsDotNet.Compiler.ExpressionBuilder.ConvertTokensToExpressions(IEnumerable`1 tokens)
   at HandlebarsDotNet.Compiler.HandlebarsCompiler.Compile(TextReader source)
   at HandlebarsDotNet.Handlebars.HandlebarsEnvironment.Compile(TextReader template)
   at HandlebarsDotNet.Handlebars.HandlebarsEnvironment.Compile(String template)
   at HandlebarsDotNet.Handlebars.Compile(String template)
   at HandlebarsDotNet.Test.IteratorTests.EmptyElementTemplate() in C:\Development\temp\Handlebars.Net\source\Handlebars.Test\IteratorTests.cs:line 31
```

The problem is that when the `Expression.Block()` is created that represents processing of the different tokens inside the iterator, there are no statements to include in the block. `Expression.Block()` fails with the exception above if the block is empty in this way.

I've worked around this by providing a no-op statement in place of the body - I'm not familiar enough with the codebase to tell whether this is reasonable, or if a fix at another layer would be more appropriate, but the tests are passing and the performance impact in the normal case should be only a comparison between the body statement count and zero.

Let me know if another fix would be more appropriate, happy to dig in :-) Cheers!

### DIFF
--- a/source/Handlebars.Test/IteratorTests.cs
+++ b/source/Handlebars.Test/IteratorTests.cs
@@ -25,6 +25,26 @@ namespace HandlebarsDotNet.Test
         }
 
         [Fact]
+        public void EmptyElementTemplate()
+        {
+            var source = "Hello,{{#each people}}{{/each}}";
+            var template = Handlebars.Compile(source);
+            var data = new
+            {
+                people = new[]{
+                    new {
+                        name = "Erik"
+                    },
+                    new {
+                        name = "Helen"
+                    }
+                }
+            };
+            var result = template(data);
+            Assert.Equal("Hello,", result);
+        }
+
+        [Fact]
         public void WithIndex()
         {
             var source = "Hello,{{#each people}}\n{{@index}}. {{name}}{{/each}}";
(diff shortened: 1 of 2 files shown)


rexm:
Released 1.9.4 to nuget

nblumhardt:
Thanks @rexm! :+1:

PR #252, opened 2018-06-04
Fix null literal binding

mohd-akram:


mohd-akram:
Since C# has no concept of undefined, I figured it would be the same as null.

rexm:
That may create some more subtle bugs down the line. Suggest we introduce an Undefined type with an equality operator overload for comparisons to null, and add a check for it in the Truthy/Falsy helpers.

mohd-akram:
Yeah, true. I've changed the PR to just check for `null` which is more important. Right now `undefined` results in `UndefinedBindingResult` which is pretty close and with #244 it's easy to check.

rexm:
Thanks! Will look at releasing this weekend.

PR #254, opened 2018-06-05
Prevent failures when handling empty block bodies in helpers

nblumhardt:
This fixes another instance of the "empty block body" problem addressed by #250 - in this case, when using helpers for conditionals - `{{#myIf}}{{else}}{{/myIf}}`.

The failure here is identical to the one covered in the other PR.


rexm:
Will look at releasing this weekend. Thanks for the contribution!

PR #259, opened 2018-07-06
Fixed UpperCamelCaseInputModelNaming() Test Case

sganz:
Updated the test case from person.address.homeCountry (lower Camel) to person.address.HomeCountry (upper Camel).


sganz:
And a huge thanks to you for the C# version!

PR #261, opened 2018-08-17
Add support for sub-expressions and refactor ExpressionVisitors

### LINKED ISSUE #208: Can't use subexpression in partial parameter
So I'm trying to do something like this:

`{{> head title=(_ "Sign Up") }}`

where `_` is a helper I've registered, but I'm getting this error:

`HandlebarsDotNet.HandlebarsCompilerException: Token ')' could not be converted to an expression`

This works in the JavaScript library:

```javascript
Handlebars.registerHelper('_', function(s) {
	return new Handlebars.SafeString(s);
});
Handlebars.registerPartial('head', 'title is "{{title}}"');
var template = Handlebars.compile('{{> head title=(_ "Sign Up") }}');
alert(template());
```

pmccloghrylaing:
Closes #208 

Refactored HandlebarsExpressionVisitor to match ExpressionVisitor implementation and simplify sub-classes.


source/Handlebars/Compiler/Translation/Expression/PartialBinder.cs
```diff
@@ -22,6 +19,11 @@ private PartialBinder(CompilationContext context)
         {
         }
 
+        protected override Expression VisitBlockHelperExpression(BlockHelperExpression bhex)
+        {
+            return bhex;
+        }
+
```
pmccloghrylaing: For some reason the PartialBinder shouldn't check within BlockHelpers or issues occur with `@partial-block`

rexm:
https://www.igvita.com/2011/12/19/dont-push-your-pull-requests/

pmccloghrylaing:
Point taken! I'll write some thoughts up in #208.

PR #262, opened 2018-08-28
Add support for parameters with sub-expressions

pmccloghrylaing:
Fix for #208


rexm:
In the expression tree, I'm not clear on the function of the `HashParameterAssignmentExpression` - it only has a `Name` property, but normally in an AST an assignment would have a target and a value. Can you clarify on how this works?

pmccloghrylaing:
It might be more accurate to specify that it's an intermediate expression, it's only the first half (`name=`). The accumulator looks for it and combines it with the following expression. It's the cleanest way I could work out to do it and handle sub expressions nested in parameters and parameters nested in sub expressions.

rexm:
Thanks. This is a pretty big change so it took a little bit to grok. Will aim to get it published this weekend.

36PopTarts:
Hi. Sorry to ask this directly in the issue, but this sounds like a change that would really help me out in the application I'm working on and it doesn't seem to have been published yet. Is there a timeframe on this one?

PR #277, opened 2018-12-18
Update Add unit tests task

jawn:
They have been added since this README was created.


rexm:
Some unit tests exist, but this task is outstanding until we have 100% coverage.

jawn:
Thanks for the clarification

PR #279, opened 2019-01-07
Add Block helpers to readme

stephengeorgewest-navitaire:
It took a while to figure out block helpers ( I found the answer by reading source/Handlebars/BuiltinHelpers.cs ). It may be helpful to have a bit about it in the read me.


rexm:
Thanks!

PR #281, opened 2019-01-15
Master

leniency:
Added two features for partials:
  - Abstracted the missing template loader to allow a custom handler. The existing file system loader is still the default option, but users could define their own.
  - Added an optional handler for when templates fail to be found. Allows graceful and/or silent handling rather than throwing a full exception.


rexm:
Thanks! I'll include these in the next rev

PR #289, opened 2019-03-14
Added support for Raw Blocks / Quad Mustache syntax

### LINKED ISSUE #179: How can I achieve a raw block helper?
I'm trying to compile a handlebars template server side that has a handlebars template in it that I want to compile clients side. The Handlebars documentation has this:

[http://handlebarsjs.com/block_helpers.html](http://handlebarsjs.com/block_helpers.html)
Raw Blocks

Raw blocks are available for templates needing to handle unprocessed mustache blocks.
```
{{{{raw-helper}}}}
  {{bar}}
{{{{/raw-helper}}}}
```
will execute the helper raw-helper without interpreting the content.
```
Handlebars.registerHelper('raw-helper', function(options) {
  return options.fn();
});
```
will render

```
{{bar}}
```

However, my attempt using Handlebars.Net still seems to be compiling the inner block. 
```
{{{#raw-helper}}}
      {{myhelper key}}
{{{/raw-helper}}}
```
```
Handlebars.RegisterHelper("raw-helper", (writer, options, context, parameters) =>{
                options.Template(writer, null);
            });
```
And so I'm getting an error that the helper myhelper is undefined, which is true for this pass since myhelper is defined client side.

If I use {{{{raw-helper}}}} I get the error: "Reached unparseable token in expression: {raw-helper}}}}"

Note: I'm using Handlebars.Net version 1.7.1, I'd prefer not to update yet, but will if it will help get me to where I want to be.

vzwick:
Fixes #179, feedback is much appreciated

### DIFF
--- a/source/Handlebars.Test/RawHelperTests.cs
+++ b/source/Handlebars.Test/RawHelperTests.cs
@@ -1,0 +1,133 @@
+﻿using System.Collections.Generic;
+using Xunit;
+
+namespace HandlebarsDotNet.Test
+{
+    public class RawHelperTests
+    {
+        [Fact]
+        public void RawBlockHelper()
+        {
+            var inst = Handlebars.Create();
+            var source = "{{{{rawBlockHelper}}}} {{foo}} {{{foo}}}{{{{otherRawBlockHelper}}}} {{ bar }}{{{bar}}}{{{{/otherRawBlockHelper}}}}{{{{/rawBlockHelper}}}}";
+
+            inst.RegisterHelper("rawBlockHelper", (writer, options, context, arguments) => {
+                options.Template(writer, null);
+            });
+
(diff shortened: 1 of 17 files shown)


vzwick:
Sorry about pushing more commits - I had blatantly overlooked a few edge cases around reassembling the body after tokenization.

rexm:
Need to check why CI isn’t running on this PR

vzwick:
[It actually is](https://travis-ci.org/rexm/Handlebars.Net/builds/506457319), just not updating the status on GitHub.

rexm:
Do you need to do any more edge case testing?

vzwick:
I don't think so - at least I didn't stumble upon any other readily apparent issues.

PR #310, opened 2019-08-02
Try to fix CallSite memory leak

DmitryDem:


rexm:
Did you profile the memory and execution time for this build vs released under load?

source/Handlebars/DynamicMetaObjectHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Runtime.CompilerServices;
+using Microsoft.CSharp.RuntimeBinder;
+
+namespace Handlebars
+{
+    internal static class DynamicMetaObjectHelper
+    {
+        internal static object GetProperty(IDynamicMetaObjectProvider target, string name)
+        {
+            var self = (Expression)Expression.Constant(target);
```
304NotModified: I think this is an unneeded cast?

DmitryDem:
> Did you profile the memory and execution time for this build vs released under load?
Hi. I will try to find free time to check memory consumption and execution time on my local machine but it's will be not be so easy, because my local machine don't so powerful as production machine (last week I wrote that memory leak occurred in highload with hundred thousands messages processing per second)
Also I thought about solution in this pull request and it's not good solution for some reasons. The first reason that we are calling inner Expression of DynamicObject for are getting property value directly bypassing CallSite - internal CallSite caching logic will not be used . The second reason that memory leak occurred just only in the main .NET framework. In the .NET Core memory leak was fixed -https://github.com/dotnet/corefx/pull/15397.
Could we write solution that just enable getting property value directly like this: if (Handlebars.Settings.DoNotUseCallSite) { logic for getting property value directly } else {logic for getting property value through CallSite}? By default option DoNotUseCallSite disabled.

source/Handlebars/DynamicMetaObjectHelper.cs
DmitryDem: Yes, because ConstantExpression inherits from Expression. I will remove unneeded cast. Thanks.

rexm:
There was additional discussion about not compiling expression trees on-the-fly in the issue thread. That will also lead to a memory leak, since each compilation represents a new assembly loaded into the AppDomain that will never be unloaded.

DmitryDem:
> There was additional discussion about not compiling expression trees on-the-fly in the issue thread. That will also lead to a memory leak, since each compilation represents a new assembly loaded into the AppDomain that will never be unloaded.

DmitryDem:
After workaround publication
image

304NotModified:
This is after JObjects to nested dictionaries? Maybe we could do something in this library for it? (Although it doesn't know JObject I guess)

rexm:
Check out the performance section of the readme.

304NotModified:
Yes I saw it already (see also https://github.com/JamesNK/Newtonsoft.Json/pull/2113), but good to know if this is the same issue

PR #313, opened 2019-08-02
Added tests and improved allocations for HtmlEncoder

304NotModified:
As the HtmlEncoder is an important thing when rendering, I added some tests for the coverage and improved the allocations, no need to allocate a stringbuilder and new string if no encoding needs to be done :)

### DIFF
--- a/source/Handlebars/HtmlEncoder.cs
+++ b/source/Handlebars/HtmlEncoder.cs
@@ -8,7 +8,7 @@ namespace HandlebarsDotNet
     {
         public string Encode(string text)
         {
-            if (text == null)
+            if (string.IsNullOrEmpty(text))
                 return String.Empty;
 
 
--- a/source/Handlebars/HtmlEncoder.cs
+++ b/source/Handlebars/HtmlEncoder.cs
@@ -11,9 +11,36 @@ namespace HandlebarsDotNet
             if (text == null)
                 return String.Empty;
 
-            var sb = new StringBuilder(text.Length);
 
+            // Detect if we need to allocate a stringbuilder and new string
             for (var i = 0; i < text.Length; i++)
+            {
+                switch (text[i])
+                {
+                    case '"':
+                    case '&':
+                    case '<':
+                    case '>':
+                        return ReallyEncode(text, i);
+                    default:
+                        if (text[i] > 159)
+                        {
+                            return ReallyEncode(text, i);
+                        }
+                        else
+
+                            break;
+                }
+            }
+
+            return text;
+        }
+
+        private static string ReallyEncode(string text, int i)
+        {
+            var sb = new StringBuilder(text.Length + 5);
+            sb.Append(text, 0, i);
+            for (; i < text.Length; i++)
             {
                 switch (text[i])
                 {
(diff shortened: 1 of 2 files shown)


304NotModified:
> [xUnit.net 00:00:01.7340644]     Handlebars.Test.HtmlEncoderTests.EncodeTest(input: "�", expected: "&#252;") [FAIL]
985[xUnit.net 00:00:01.7409369]       Assert.Equal() Failure
986[xUnit.net 00:00:01.7421150]                   ↓ (pos 2)
987[xUnit.net 00:00:01.7431782]       Expected: &#252;
988[xUnit.net 00:00:01.7441620]       Actual:   &#65533;

That's strange, it works local :|

rexm:
Thanks for the additional test. It's not clear to me how the allocation change is a net improvement.

304NotModified:
Allocations are taking memory and makes the GC busy. 

FYI: We've done a lot of those improvements in NLog.

304NotModified:
> That's strange, it works local :|

This is fixed with the UTF-8 BOM on the file

rexm:
Yeah I get that generally speaking allocations take memory and make GC busy, but in this specific case it is not clear to me what the net improvement is.

304NotModified:
less memory usage and lower GC pressure for all renderings?

rexm:
Great! How much?

304NotModified:
Dunno, depends on the machine. Could try to measure

rexm:
Can always leave the unit tests and remove the change. Generally we don't put in changes that  claim to be performance improvements without instrumenting them.

304NotModified:
> Generally we don't put in changes that claim to be performance improvements without instrumenting them.

That's clear ;)


> Can always leave the unit tests and remove the change.

If I don't find time this week, then that's OK. I think it could do the performance instrumention with https://www.jetbrains.com/dotmemory/unit/

304NotModified:
I tested this case:

```c#
        [Fact]
        public void LargeEncodeTest()
        {
            var input = new string('0', 1_000_000_000);
            // Arrange
            var htmlEncoder = new HtmlEncoder();


            // Act
            var result = htmlEncoder.Encode(input);

        }
```

before:

image

after:

image

all run 3 times, release mode.

304NotModified:
before:

image


after:

image

rexm:
Thanks! New release probably this weekend.

304NotModified:
Great!

source/Handlebars/HtmlEncoder.cs
```diff
@@ -11,9 +11,36 @@ public string Encode(string text)
...
+                            break;
+                }
+            }
+
+            return text;
+        }
+
+        private static string ReallyEncode(string text, int i)
+        {
+            var sb = new StringBuilder(text.Length + 5);
```
roydukkey: @304NotModified Why are you adding 5 to the length? How does this help during allocation?

source/Handlebars/HtmlEncoder.cs
```diff
@@ -11,9 +11,36 @@ public string Encode(string text)
...
+                    case '&':
+                    case '<':
+                    case '>':
+                        return ReallyEncode(text, i);
+                    default:
+                        if (text[i] > 159)
+                        {
+                            return ReallyEncode(text, i);
+                        }
+                        else
```
roydukkey: @304NotModified Could this else be omitted? Or, at the very least be use braces `{}` to match the style of the rest of the project?

source/Handlebars/HtmlEncoder.cs
```diff
@@ -34,14 +61,16 @@ public string Encode(string text)
                         if (text[i] > 159)
                         {
                             sb.Append("&#");
-                            sb.Append(((int) text[i]).ToString(CultureInfo.InvariantCulture));
+                            sb.Append(((int)text[i]).ToString(CultureInfo.InvariantCulture));
                             sb.Append(";");
                         }
                         else
```
roydukkey: @304NotModified I guess I see the braces existed like this before your edits. So maybe this would be an appropriate time to fix?

PR #315, opened 2019-08-05
 Added support for implicit interface implementation for dictionaries

304NotModified:
Fixes https://github.com/rexm/Handlebars.Net/issues/239


source/Handlebars.Test/BasicIntegrationTests.cs
```diff
@@ -15,7 +15,8 @@ public void BasicPath()
         {
             var source = "Hello, {{name}}!";
             var template = Handlebars.Compile(source);
-            var data = new {
+            var data = new
```
304NotModified: FYI, I used "format document". Not sure if this is a problem

rexm:
Thanks!

PR #319, opened 2019-08-07
Literals should be the correct type when passed as arguments to helpers

### LINKED ISSUE #298: Handlebars 1.10.1 : RegisterHelper different?
I register this handlebars custom code:
``` c#
handlebarsContext.RegisterHelper("Random", (writer, context, arguments) =>
{
});
```
And when processing this text:
`{{Random Type=\"Text\" Min=8 Max=20}}`

The arguments look like this:
image

Note that the **Min** and **Max** are defined as a string instead of a int ?

Is this something which has been changed lately?

pmccloghrylaing:
Closes #298

### DIFF
--- a/source/Handlebars.Test/HelperTests.cs
+++ b/source/Handlebars.Test/HelperTests.cs
@@ -424,6 +424,51 @@ namespace HandlebarsDotNet.Test
             var output = template(data);
             Assert.Equal("Inverse", output);
         }
+
+        [Fact]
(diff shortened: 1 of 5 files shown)


StefH:
Is this change already present in nuget?

StefH:
@rexm : Can you please release a version 1.x.x NuGet which includes this PR ?

nzjoel1234:
It would be great if this could be released - is there any indication when this could happen?
Thanks

Issue #243, opened 2018-01-25
@last element in nested each refers to the parent item ( C# )


ip75:
I try to dump array with objects to csv file. Each object is a row with values.

{{#unless @last}};{{/unless}} block in nested loop is interpreted as a last parent element.

There is a code for LinqPad:

    void Main()
    {
    	string template = "{{#each header}}{{this}}{{#unless @last}};{{/unless}}{{/each}}\n{{#each rows}}{{#each this}}{{this}}{{#unless @last}};{{/unless}}{{/each}}\n{{/each}}\n\n\n\n\n\n";			
    			
    			var data = new
                {
                    header = new string [] { "Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat" },
                    rows = new item [] { new item(), new item(), new item(), new item(), new item(), new item(), new item() }
    				
    			};
    
    	var handlebarsInstance = Handlebars.Create(new HandlebarsConfiguration {TextEncoder = new CsvEncoder()});
    	Console.WriteLine(handlebarsInstance.Compile(template).Invoke(data));
    }
    
    public class CsvEncoder : ITextEncoder
    {
        public string Encode(string value)
        {
            return value;
        }
    }
    
    public class item
    {
    	public string val1 {get; set;} = "A";
    	public string val2 {get; set;} = "B";
    	public string val3 {get; set;} = "C";
    	public string val4 {get; set;} = "D";
    	public string val5 {get; set;} = "E";
    	public string val6 {get; set;} = "F";
    	public string val7 {get; set;} = "G";
    }

result is:

    Sun;Mon;Tue;Wed;Thu;Fri;Sat
    A;B;C;D;E;F;G;
    A;B;C;D;E;F;G;
    A;B;C;D;E;F;G;
    A;B;C;D;E;F;G;
    A;B;C;D;E;F;G;
    A;B;C;D;E;F;G;
    ABCDEFG

I expect this result:

    Sun;Mon;Tue;Wed;Thu;Fri;Sat
    A;B;C;D;E;F;G
    A;B;C;D;E;F;G
    A;B;C;D;E;F;G
    A;B;C;D;E;F;G
    A;B;C;D;E;F;G
    A;B;C;D;E;F;G
    A;B;C;D;E;F;G

Is this error in my handlebars template or bug in handlebars parser/composer for .net?

https://stackoverflow.com/questions/48424658/last-element-in-nested-each-handlebars-c-sharp

[...]

PR #330, opened 2019-12-18
#243: Support `@last` for properties enumeration

zjklee:
The PR adds support for `@last` property for properties enumeration.
Originally raised in #243 issue.


rexm:
Thanks!

Issue #329, opened 2019-12-13
Keys thats are number-strings cannot be escaped


alexdawes:
When the data being passed to a compiled template has a field that is a number as a string, the template fails to find this field even if escaped.

```using System;
using HandlebarsDotNet;
using Newtonsoft.Json.Linq;

namespace HandlebarsTester
{
    public static class Program
    {
        internal static string Test(string template, string json)
        {
            var jObject = JObject.Parse(json);
            var compiled = Handlebars.Compile(template);
            var result = compiled.Invoke(jObject);
            return result;
        }

        public static void Main(string[] args)
        {
            Console.WriteLine(Test("123: {{ [123] }}", "{ \"123\": \"FOO\" }"));
            Console.ReadLine();
        }
    }
}
```

The above code prints: ```123: ```. Doing the equivalent in handlebarsjs returns ```123: FOO```

thomasraynal:
try this :  Console.WriteLine(Test("123: {{ [[123]] }}", "{ \"123\": \"FOO\" }"));

see: https://github.com/rexm/Handlebars.Net/issues/325

alexdawes:
Having looked further, it appears that the issue is specific to `JObject`s. The problem is that `JObject` is assignable to `IEnumerable<object>`, and this then triggers the enumerable indexing code path, instead of the object one. In my example above, because the `JObject` doesn't have 123rd item, this then returns an `UndefinedBindingResult`, but if I had used `{{ [0] }}` as a template instead, then it would have returned `"123": "FOO"`, as the 0th item is a `JProperty` with that as the ToString representation.

I'm working on an attempt to fix this at the moment.

alexdawes:
> try this : Console.WriteLine(Test("123: {{ [[123]] }}", "{ "123": "FOO" }"));
> 
> interestingly enought, this work too : Console.WriteLine(Test("123: {{ [123]] }}", "{ "123": "FOO" }"));
> 
> see: #325

FYI this is because the key is trimmed using `.Trim('[', ']')`, which is probably a bit heavy handed. The PR I have just opened changes this to only trim one of each at most, and only if they both exist (at the start and end of the keys respectively)

Issue #328, opened 2019-12-13
Escaping of keys does not work with . characters


alexdawes:
When a key has a `.` in, handlebars.net fails to find the key, even when escaping with `[...]`

```
using System;
using HandlebarsDotNet;
using Newtonsoft.Json.Linq;

namespace HandlebarsTester
{
    public static class Program
    {
        internal static string Test(string template, string json)
        {
            var jObject = JObject.Parse(json);
            var compiled = Handlebars.Compile(template);
            var result = compiled.Invoke(jObject);
            return result;
        }

        public static void Main(string[] args)
        {
            Console.WriteLine(Test("foo.bar: {{ [foo.bar] }}", "{ \"foo.bar\": \"FOO\" }"));
            Console.ReadLine();
        }
    }
}
```

The above example prints: ```foo.bar: ```. The equivalent using handlebarsjs successfully renders this as ```foo.bar: FOO```

PR #333, opened 2019-12-20
Minor improvements to path handling and access

alexdawes:
Ensure that path breaking respects square bracket escape characters so that objects with keys like `foo.bar` can be accessed using `{{ [foo.bar] }}` (#328)

Give preference to string index property getters if one exists, to avoid classes that implement `IEnumerable<object>` whilst also having a string indexer (most notably `JObject`) from being unable to be used with numeric keys. (#329)

Only trim a single pair of square brackets from the key, and only if both exist (`[` at start of key and `]` at end)


alexdawes:
#329 #328

Issue #301, opened 2019-05-26
Sequence contains more than one element


johnknoop:
I have a Handlebars template that I've verified works both on http://tryhandlebarsjs.com/ and using the Handlebars Preview extension in VS Code.

However, I get this exception when compiling it with Handlebars.Net:

```csharp
System.InvalidOperationException: Sequence contains more than one element
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source)
   at HandlebarsDotNet.Compiler.IteratorBlockAccumulatorContext.IsClosingElement(Expression item)
   at HandlebarsDotNet.Compiler.BlockAccumulator.AccumulateBlock(IEnumerator`1 enumerator, BlockAccumulatorContext context)
   at HandlebarsDotNet.Compiler.BlockAccumulator.AccumulateBlock(IEnumerator`1 enumerator, BlockAccumulatorContext context)
   at HandlebarsDotNet.Compiler.BlockAccumulator.AccumulateBlock(IEnumerator`1 enumerator, BlockAccumulatorContext context)
   at HandlebarsDotNet.Compiler.BlockAccumulator.ConvertTokens(IEnumerable`1 sequence)+MoveNext()
   at System.Collections.Generic.List`1.AddEnumerable(IEnumerable`1 enumerable)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at HandlebarsDotNet.Compiler.ExpressionBuilder.ConvertTokensToExpressions(IEnumerable`1 tokens)
   at HandlebarsDotNet.Compiler.HandlebarsCompiler.Compile(TextReader source)
   at HandlebarsDotNet.Handlebars.HandlebarsEnvironment.Compile(String template)
```

This is the template:

```handlebars
<div>
{{#each Items as |item|}}
	<span>
	{{#if item.HasVariants}}
		<strong>
		{{#each item.Variants as |variant|}}
			<i>{{variant.Code}}</i>
		{{/each}}
		</strong>
	{{else}}
		<strong>
		{{item.Code}}
		</strong>
	{{/if}}
	</span>
{{/each}}
</div>
```

Any ideas?

rexm:
The `as |alias|` syntax isn't supported yet.

johnknoop:
I see. Thanks!

PR #335, opened 2020-01-12
Support for BlockParams

zjklee:
The PR introduces support of BlockParams syntax (#301, mentioned in #294): 
```{{#each enumerateMe as |item id|}}{{@item}}: {{@id}} {{/each}}```
P.S. supports both `@item` and `item` syntax for referring param.

The implementation was guided by emberjs/rfcs/pull/3

Where supported:
- #each:
  - object iterators: first param maps to `@value`, second to `@key`
  - enumerable iterators: first param maps to `@value`, second to `@index`
- #with
- custom helpers via configuring with call to `options.BlockParams(...)`

Notes:
Implementation done in a way to potentially support more keywords then just `as` (in case more would be added in the future) but [strictly limited](https://github.com/rexm/Handlebars.Net/compare/master...zjklee:blockParams-support?expand=1#diff-2b14f2a9d6500f9d13b324ee9b9acd04R31) to support only it for now.

### DIFF
--- a/source/Handlebars.Test/BasicIntegrationTests.cs
+++ b/source/Handlebars.Test/BasicIntegrationTests.cs
@@ -144,18 +144,9 @@ false
                 }
             };
 
-            try
-            {
-                template(data);
-            }
-            catch (HandlebarsUndefinedBindingException ex)
-            {
-                Assert.Equal("person.lastname", ex.Path);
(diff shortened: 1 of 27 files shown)


zjklee:
May be ignored in case rexm/Handlebars.Net/pull/352 is merged.

PR #348, opened 2020-05-27
Added Tests for use in discoverring issue with escape characters

mcgear:
As i was testing some things we are trying to do with Handlebars.net in some json files and ran into an issue, when compiling a template with, the use of '\\'.  I added two unit tests, one to show Json serialization working without escape sequences, the other to demonstrate it broken.

The issue does not seem to happen against other escape sequences like \" as you'll see in the unit test.


mcgear:
I was able to get some time to take a look at the issue, and i believe that the simple two liner that i added allows for proper node management in a case where multiple \\ is necessary to remain in the template.

All original Unit Tests are passing as well as the new ones for this specific case.

PR #403, opened 2020-12-10
Multiple fixes to path resolvement

zjklee:
What's inside:
- `@data` flows to child context
- helper names are case insensitive
- fixed context switch
- fixed problem with fallback to context when same name variable exists

Issues:
- found in /issues/394
- closes /issues/402
- closes /issues/401
- closes /issues/399

### DIFF
--- a/source/Handlebars.Test/IssueTests.cs
+++ b/source/Handlebars.Test/IssueTests.cs
@@ -1,6 +1,10 @@
 using System;
 using System.Dynamic;
 using System.IO;
+using HandlebarsDotNet.Collections;
+using HandlebarsDotNet.Helpers;
+using HandlebarsDotNet.PathStructure;
+using HandlebarsDotNet.StringUtils;
 using Xunit;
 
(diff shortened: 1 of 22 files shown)


source/Handlebars.Test/IssueTests.cs
```diff
@@ -220,5 +224,97 @@ public void SlashesInTemplateKey()
...
+            const string expected = "<ul><li>a 1 zzz z</li><li>b 2 yyy y</li><li>c 3 xxx x</li></ul>";
+            var actual = result
+                .Replace("\n", string.Empty)
+                .Replace("\r", string.Empty)
+                .Replace("\t", string.Empty);
+            
+            Assert.Equal(expected, actual);
+        }
+        
+        private class JoinHelper : IHelperDescriptor<HelperOptions>
```
zjklee, line 287: @StefH 
Some inspiration on allocation-less implementation of Join helper based on your example in Handlebars-Net/Handlebars.Net/issues/399

Issue #412, opened 2021-01-07
Accessing struct properties throws exception in ReflectionMemberAccessor


leniency:
### Describe the bug
Accessing a struct property causes an ArgumentException in ReflectionMemberAccessor.CreateGetDelegate

`Cannot bind to the target method because its signature or security transparency is not compatible with that of the delegate type.'`

### Expected behavior:
Struct properties should be accessible like class properties.

### Test to reproduce
```csharp
[Fact]
        public void StructReflectionAccessor()
        {
            const string template = "{{Name}}";

            var handlebars = Handlebars.Create();
            var handlebarsTemplate = handlebars.Compile(template);

            var actual = handlebarsTemplate(new CustomStruct
            {
                Name = "Foo"
            });

            Assert.Equal("Foo", actual);
        }

        private struct CustomStruct
        {
            public string Name { get; set; }
        }
```

PR #413, opened 2021-01-07
Fix for Issue #412

leniency:
Unit test and fix for issue #412 - The ReflectionMemberAccessor throws an exception when accessing value type members. The internal CreateGetDelegate method must specify pass-by-reference for these types.

### DIFF
--- a/source/Handlebars/MemberAccessors/ReflectionMemberAccessor.cs
+++ b/source/Handlebars/MemberAccessors/ReflectionMemberAccessor.cs
@@ -1,11 +1,11 @@
-using HandlebarsDotNet.Collections;
-using HandlebarsDotNet.EqualityComparers;
-using HandlebarsDotNet.PathStructure;
-using HandlebarsDotNet.Runtime;
 using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


source/Handlebars/MemberAccessors/ReflectionMemberAccessor.cs
```diff
@@ -92,20 +92,37 @@ public RawObjectTypeDescriptor(Type type)
...
                 if (field != null)
                 {
                     return o => field.GetValue(o);
                 }
 
                 return null;
-            }
+            }
+
+            delegate TValue ValueTypeGetterDelegate<T, TValue>(ref T instance);
```
zjklee, line 105: nit: Consider moving delegate type declaration to the beginning of the class and declare it explicitly `private`.

zjklee (approved):

leniency:
Arg, sorry for the last commit - wasn't paying attention and missed that Visual Studio updated the sln file and also normalized all the line endings in that last commit. I'll see if I can get it removed and just the delegate line fixed.

zjklee:
I've finally fixed broken Github Actions. Can you please rebase your branch on top of the latest master?

PR #443, opened 2021-03-25
Render layouts using LayoutViewModel instead of DynamicViewModel

### LINKED ISSUE #440: `Dictionary<string, object>` or `dynamic` model is not supported by layout views
### Describe the bug
When the `context` parameter contains a model of type `Dictionary<string, object>` or `dynamic`, expressions in a layout view cannot be bound to it.

This seems to be because the way context is passed to layout views is by wrapping the output of the ‘main’ view and the original context into a `DynamicViewModel`. Unfortunately, `DynamicViewModel` only supports reflection member access.

This conflicts with the guidance in the documentation that, for the best performance, models should be dictionaries.

### Expected behavior:
When the value of the `context` parameter of a view is a `Dictionary<string, object>` or a `dynamic`, I expect that expressions in the layout view are able to resolve values as they do in regular views.

In more detail: I expect that `DynamicViewModel` is capable of handling regular dictionaries and dynamic values, as well as POCOs.

### Test to reproduce
```csharp
[Fact]
public void CanUseDictionaryModelInLayout()
{
    var files = new FakeFileSystem()
                {
                    { "views\\layout.hbs", "Layout: {{property}}\r\n{{{body}}}" },
                    { "views\\someview.hbs", "{{!< layout}}\r\n\r\nBody: {{property}}" },
                };

    var handlebarsConfiguration = new HandlebarsConfiguration() { FileSystem = files };
    var handlebars = Handlebars.Create(handlebarsConfiguration);
    var render = handlebars.CompileView("views\\someview.hbs");
    var output = render(
        new Dictionary<string, object>
        {
            { "property", "Foo" },
        }
    );

    Assert.Equal("Layout: Foo\r\n\r\nBody: Foo", output);
    // actual value is "Layout: \r\n\r\nBody: Foo"
}

[Fact]
public void CanUseDynamicModelInLayout()
{
    var files = new FakeFileSystem()
                {
                    { "views\\layout.hbs", "Layout: {{property}}\r\n{{{body}}}" },
                    { "views\\someview.hbs", "{{!< layout}}\r\n\r\nBody: {{property}}" },
                };

    dynamic model = new MyDynamicModel();
    var handlebarsConfiguration = new HandlebarsConfiguration() { FileSystem = files };
    var handlebars = Handlebars.Create(handlebarsConfiguration);
    var render = handlebars.CompileView("views\\someview.hbs");
    var output = render(
        model
    );

    Assert.Equal("Layout: Foo\r\n\r\nBody: Foo", output);
    // actual value is "Layout: \r\n\r\nBody: Foo"
}

private class MyDynamicModel: DynamicObject
{
    private readonly Dictionary<string, object> _properties =
        new Dictionary<string, object>
        {
            { "property", "Foo" },
        };

    public override IEnumerable<string> GetDynamicMemberNames() => _properties.Keys;
    public override bool TryGetMember(GetMemberBinder binder, out object result) =>
        _properties.TryGetValue(binder.Name, out result);
}
```

heemskerkerik:
Fixes #440.

Because `DynamicViewModel` is ‘too flexible’, it is difficult to provide a fully working `ObjectDescriptor` for it. Instead, this introduces a new, more limited `LayoutViewModel`, which intercepts binding expressions for `body`, but forwards everything else, including iteration, to the main view’s model. `DynamicViewModel` is left unmodified, since it’s public and it’s unclear whether people actually use it.

### DIFF
--- a/source/Handlebars.Test/ViewEngine/ViewEngineTests.cs
+++ b/source/Handlebars.Test/ViewEngine/ViewEngineTests.cs
@@ -190,6 +190,29 @@ namespace HandlebarsDotNet.Test.ViewEngine
         }
 
         [Fact]
+        public void CanBindToModelInNestedLayout()
+        {
+            var files = new FakeFileSystem
+                        {
+                            { "views\\parent_layout.hbs", "Parent layout: {{property}}\r\n{{{body}}}" },
+                            { "views\\layout.hbs", "{{!< parent_layout}}\r\nLayout: {{property}}\r\n{{{body}}}" },
+                            { "views\\someview.hbs", "{{!< layout}}\r\n\r\nBody: {{property}}" },
+                        };
+
+            var handlebarsConfiguration = new HandlebarsConfiguration { FileSystem = files };
+            var handlebars = Handlebars.Create(handlebarsConfiguration);
+            var render = handlebars.CompileView("views\\someview.hbs");
+            var output = render(
+                new
+                {
+                    property = "Foo"
+                }
+            );
+
+            Assert.Equal("Parent layout: Foo\r\nLayout: Foo\r\n\r\nBody: Foo", output);
+        }
+
+        [Fact]
         public void CanUseNullModelInLayout()
         {
             var files = new FakeFileSystem
--- a/source/Handlebars.Test/ViewEngine/ViewEngineTests.cs
+++ b/source/Handlebars.Test/ViewEngine/ViewEngineTests.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Dynamic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using Xunit;
 
@@ -50,7 +52,7 @@ namespace HandlebarsDotNet.Test.ViewEngine
             //Then the correct output should be rendered
             Assert.Equal("layout start\r\nThis is the body\r\nlayout end", output);
         }
-        
+
         [Fact]
         public void CanLoadAViewWithALayoutInTheRoot()
         {
@@ -63,7 +65,7 @@ namespace HandlebarsDotNet.Test.ViewEngine
             };
 
             //When a viewengine renders that view
-            var handlebars = HandlebarsDotNet.Handlebars.Create(new HandlebarsConfiguration() {FileSystem = files});
+            var handlebars = Handlebars.Create(new HandlebarsConfiguration() {FileSystem = files});
             var render = handlebars.CompileView("views\\someview.hbs");
             var output = render(null);
 
@@ -110,7 +112,7 @@ namespace HandlebarsDotNet.Test.ViewEngine
             //Then the correct output should be rendered
             Assert.Equal("layout start\r\nThis is the body\r\nlayout end", output);
         }
-        
+
         [Fact]
         public void CanRenderAGlobalVariable()
         {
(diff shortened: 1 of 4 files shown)


source/Handlebars.Test/ViewEngine/ViewEngineTests.cs
```diff
@@ -147,7 +149,212 @@ public void CanIgnoreCommentsContainingHtml()
...
+                            { "views\\layout.hbs", "Layout: {{#each this}}{{#if @First}}First:{{/if}}{{#if @Last}}Last:{{/if}}{{this}};{{/each}}{{{body}}}" },
+                            { "views\\someview.hbs", "{{!< layout}} View" },
+                        };
+
+            var handlebarsConfiguration = new HandlebarsConfiguration
+                                          {
+                                              FileSystem = files,
+                                              Compatibility =
+                                              {
+                                                  SupportLastInObjectIterations = true,
```
zjklee, line 308: nit: This flag is not used here

source/Handlebars.Test/ViewEngine/ViewEngineTests.cs
```diff
@@ -147,7 +149,212 @@ public void CanIgnoreCommentsContainingHtml()
...
```
zjklee, line 334: nit: This flag is not used here

source/Handlebars/LayoutViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+using HandlebarsDotNet.Iterators;
+using HandlebarsDotNet.MemberAccessors;
+using HandlebarsDotNet.ObjectDescriptors;
+using HandlebarsDotNet.PathStructure;
+
+namespace HandlebarsDotNet
+{
+    internal class LayoutViewModel
+    {
+        private static readonly ChainSegment _bodyChainSegment = ChainSegment.Create("body");
```
zjklee, line 14: Rename to `BodyChainSegment`: `static readonly` fields should be named starting with capital letter.
Add empty lime between `static` and `instance` members.

source/Handlebars/LayoutViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+        public LayoutViewModel(string body, object value)
+        {
+            _body = body;
+            _value = value;
+            _valueDescriptor = ObjectDescriptor.Create(value);
+        }
+
+        internal class DescriptorProvider: IObjectDescriptorProvider
+        {
+            private static readonly object[] _bodyProperties = { _bodyChainSegment };
```
zjklee, line 28: Rename to `BodyProperties`: `static readonly` fields should be named starting with capital letter.

source/Handlebars/LayoutViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+            private static readonly Type _type = typeof(LayoutViewModel);
```
zjklee, line 29: Rename to `Type`: `static readonly` fields should be named starting with capital letter.

source/Handlebars/LayoutViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+            private readonly ObjectDescriptor _descriptor;
```
zjklee, line 30: Add empty lime between `static` and `instance` members.

zjklee (changes requested):

source/Handlebars.Test/ViewEngine/ViewEngineTests.cs
heemskerkerik: I actually 'needed' to do this, because otherwise `{{@Last}}` is never true. Because of this: https://github.com/Handlebars-Net/Handlebars.Net/blob/master/source/Handlebars/ValueProviders/ObjectIteratorValues.cs#L66

zjklee: But in this case you're iterating over `IEnumerable`, don't you?

heemskerkerik: You're absolutely correct. I blindly copied this from the other tests, assuming I would also need it there. Fixed.

zjklee (approved):

Issue #447, opened 2021-04-27
Exception thrown for immutable array


wrong42:
### Describe the bug
Calling `HandlebarsTemplate` fails when context contains data of type `ImmutableArray<T>`.

### Expected behavior:
Immutable arrays (structs in general) are supported as valid data types in context for template.

### Test to reproduce
```csharp
[Fact]
public void ImmutableArrayTest()
{
    var itemsList = new List<string> { "hello", "world" };
    var items = ImmutableArray.CreateRange(itemsList);

    string source =
        @"{{#each Items}}
            {{this}}
          {{/each}}";

    var template = Handlebars.Compile(source);
    var result = template(new { Items = items }); // This throws exception
}
```

### Other related info
Immutable arrays were supported in version 1.1.15. After upgrading to 2.0.7 calling the `HandlebarsTemplate` throws `System.ArgumentException` with message: "Method HandlebarsDotNet.ObjectDescriptors.EnumerableObjectDescriptor.ListObjectDescriptorFactory: type argument 'System.Collections.Immutable.ImmutableArray`1[System.Uri]' violates the constraint of type parameter 'T'."

---

### After submitting the issue
Please consider contributing to the project by submitting a PR with a fix to the issue. 
This would help to solve your problem in a shorter time as well as help other users of the project.

In case you do not know where to start - feel free to ask for help in the issue thread.

Building an active community is essential for any project survival as time of maintainers is limited.

bsagal:
I believe this bug is related to the fact that `ImmutableArray<T>` is a `struct`.

The  `EnumerableObjectDescriptor .ListObjectDescriptorFactory<T, TValue>` [method ](https://github.com/Handlebars-Net/Handlebars.Net/blob/01ad0e69e56331371f85a1c59c008d7ecfd06f78/source/Handlebars/ObjectDescriptors/EnumerableObjectDescriptor.cs#L131)(and similar methods for other collections) has a genetic constraint of `where T : class, IList<TValue>`.

Is there a reason for the `class` constraint?

This method was added in commit 855057b.

zjklee:
Hello @bsagal 
No, I do not think there is any real reason to have a `class` constraint.

PR #449, opened 2021-05-05
Allow struct collections

bsagal:
Fix bug #447

### DIFF
--- a/source/Handlebars.Test/Handlebars.Test.csproj
+++ b/source/Handlebars.Test/Handlebars.Test.csproj
@@ -50,6 +50,7 @@
     <PackageReference Include="NSubstitute" Version="4.2.2" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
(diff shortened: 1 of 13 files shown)


zjklee:
Hello @bsagal 
Can you please add a test that covers the bug?

bsagal:
In order to test all the fixed cases I will need to create struct implementers of the various collection interfaces.

For `IList<T>` I can use the `System.Collections.Immutable.ImmutableArray<T>`, however this will only test one problematic case.
Additionally this will require and extra dependency on `System.Collections.Immutable`.

I am unaware of any struct implementations of all the collection interfaces included in the framwork and am unaware of a way to mock stucts.

As far as I am aware to only solution is to create struct implementations for all the collection interfaces.

Where would you suggest I place in implemations in the source tree?

zjklee:
Having an extra dependency in tests is fine. I think having coverage at least for `ImmutableArray` should be good enough at this point.

bsagal:
Ok understood.

This will help test a stuct `IList<T>`.
What should i do about testing stuct implamentations of the other collection interfaces such as `IDictionary<T>`. They were also fixed.

PR #451, opened 2021-06-01
Improve performance for large arrays (10.000+)

albertschulz:
When rendering templates with large arrays, execution time grows exponentially.
Take the following measurements:

```
Elements | Rendering Time (ms)
  20.000 |          5.500
  40.000 |         20.000
  80.000 |         80.000
```

produced by the following minimal example:

```csharp
class Program
	{
		static void Main(string[] args)
		{
			Stopwatch sw = new Stopwatch();
			
			var source = @"{{#each this}}{{this}}{{/each}}";
			var data = Enumerable.Range(0, 20000).ToList();
			
			sw.Start();
			template(data);
			sw.Stop();
			Console.WriteLine($"Filling 20.000 Elements took {sw.ElapsedMilliseconds} ms");
			
			sw.Restart();
			data = Enumerable.Range(0, 40000).ToList();
			template(data);
			sw.Stop();
			Console.WriteLine($"Filling 40.000 Elements took {sw.ElapsedMilliseconds} ms");
			
			sw.Restart();
			data = Enumerable.Range(0, 80000).ToList();
			template(data);
			sw.Stop();
			Console.WriteLine($"Filling 80.000 Elements took {sw.ElapsedMilliseconds} ms");
		}
	}
```

The issue originated from the cache for the iterators contained in `BoxedValues.cs`.
In opposite to the desired reduced memory usage, memory usage actually also explodes for the example show above.

This pull request removes the cache, leaving the small boxing-cache for integer values between 0 and 20.
Furthermore, it adds a benchmark for the execution time.
For some reasons, I could not reproduce the dramatic execution times stated here with this benchmark, although it uses the same code. Anyone has an idea, why the benchmarks execute faster than expected?

### DIFF
--- a/source/Handlebars.Benchmark/LargeArray.cs
+++ b/source/Handlebars.Benchmark/LargeArray.cs
@@ -10,27 +10,20 @@ namespace HandlebarsNet.Benchmark
     public class LargeArray
     {
         private object _data;
-        private HandlebarsTemplate<TextWriter, object, object> _default;
+        private HandlebarsTemplate<object, object> _default;
 
-        [Params(1, 2, 3, 4, 5)]
+        [Params(20000, 40000, 80000)]
         public int N { get; set; }
         
         [GlobalSetup]
         public void Setup()
         {
             const string template = @"{{#each this}}{{this}}{{/each}}";
-
-            var handlebars = Handlebars.Create();
-
-            using (var reader = new StringReader(template))
-            {
-                _default = handlebars.Compile(reader);
-            }
-
-            _data = Enumerable.Range(0, (int) Math.Pow(10, N)).ToList();
+            _default = Handlebars.Compile(template);
+            _data = Enumerable.Range(0, N).ToList();
         }
         
         [Benchmark]
-        public void Default() => _default(TextWriter.Null, _data);
+        public void Default() => _default(_data);
     }
 }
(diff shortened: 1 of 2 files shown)


zjklee:
> For some reasons, I could not reproduce the dramatic execution times stated here with this benchmark, although it uses the same code. Anyone has an idea, why the benchmarks execute faster than expected?

This might be related to the measurement approach. Benchmark would executed with the warmup. During the warmup all the values would be cache and would be reused in further iterations.
In the example above you perform a "cold" run for a subset of values each time so you were observing caching penalty.

source/Handlebars.Benchmark/LargeArray.cs
```diff
@@ -0,0 +1,29 @@
...
+        private HandlebarsTemplate<object, object> _default;
+
+        [Params(20000, 40000, 80000)]
+        public int N { get; set; }
+        
+        [GlobalSetup]
+        public void Setup()
+        {
+            const string template = @"{{#each this}}{{this}}{{/each}}";
+            _default = Handlebars.Compile(template);
```
zjklee, line 22: It's better to use `TextWriter` overload in the benchmark (like you did in `5c7fc6d27a62b21b1edb06de96e2f05844cea3e4` commit) as it allows to isolate memory allocations to internals only without noise produced by additional string allocations.

zjklee (changes requested):

source/Handlebars.Benchmark/LargeArray.cs
albertschulz: Here you go 78dea01

zjklee (approved):

PR #456, opened 2021-06-22
Fix memory leak at Configuration subscribers

### LINKED ISSUE #452: [Question] Memory concerns?
Hi all, thanks for the great library.

I am trying to diagnose what appears to be a memory leak in one of our applications, and I am wanting to make sure I am not misusing or overlooking something with Handlebars.
After running our application, which will dynamically load up Handlebars templates, for a long time, generating many templates, I see memory usage continue to grow.

There's many other pieces to this, so I didn't even consider Handlebars until I used Visual Studio's memory profiler and found this:

image

Where we have (this is a diff) over 200 MB of these ReaderWriterLockSlim objects, and they're all referenced from HandlebarsDotNet objects.

Our usage of Handlebars is as follows:

```csharp
string OnEachRequest(...)
{

    //Register custom helpers
    ...
    
    return handlebars.Compile(someTemplate)(someData);
}
```

Is there some sort of cleanup I'm forgetting to do?

Thanks! I will continue to dig into this further but figure it doesn't hurt to ask in case I'm missing something obvious.

zjklee:
What's inside:
- fixes memory leak caused by observable subscribers were not removed

Issues:
- closes #452

### DIFF
--- a/source/Handlebars/Collections/WeakCollection.cs
+++ b/source/Handlebars/Collections/WeakCollection.cs
@@ -1,0 +1,82 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace HandlebarsDotNet.Collections
+{
+    public class WeakCollection<T> : IEnumerable<T> where T : class
+    {
+        private readonly List<WeakReference<T>> _store = new List<WeakReference<T>>();
+        private int _firstAvailableIndex = 0;
+
+        public void Add(T value)
+        {
+            for (var index = _firstAvailableIndex; index < _store.Count; index++)
+            {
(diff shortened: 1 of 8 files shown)


source/Handlebars/Collections/WeakCollection.cs
```diff
@@ -0,0 +1,82 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace HandlebarsDotNet.Collections
+{
+    public class WeakCollection<T> : IEnumerable<T> where T : class
```
zjklee, line 7: Need to add tests for the class

Issue #468, opened 2021-09-27
HandlebarsDotNet.Handlebars.Configuration.NoEscape Applied Inconsistently


tonysneed:
### Describe the bug
When iterating over an array using `#each`, the value of `Configuration.NoEscape` reverts to `false`.

### Expected behavior:
`Configuration.NoEscape` remains `true` when set to `true`.

### Test to reproduce
See PR [180](https://github.com/TrackableEntities/EntityFrameworkCore.Scaffolding.Handlebars/pull/180) in my repo.

zjklee:
Hello @tonysneed 
Can you please include the minimal test to the issue here to simplify investigation?

tonysneed:
Here is a [repo](https://github.com/tonysneed/handlebars-encoding-test) with minimal tests that reproduce the issue.

tonysneed:
@zjklee Have you been able to look at the repo with minimal tests that I provided?

zjklee:
Hello @tonysneed 
Unfortunately, I had no spare time to take a look. I'd try to take a look in a next couple of days but I cannot promise when it would happen.
I'd be happy to help you with investigation if you decide to take a deeper look on your own.

PR #473, opened 2021-11-25
HtmlEncoding consistent with rules in handlebars.js

tommysor:
The original issue this PR was intended to solve have been fixed in PR #477.
This PR now deals with general rules for encoding in Handlebars.Net vs handlebars.js.

Using this PR since it contains the history of how this change came to be.

~~**[WIP] Configuration.NoEscape inconsistent fix**~~

~Regarding issue: HandlebarsDotNet.Handlebars.Configuration.NoEscape Applied Inconsistently #468~

~~Change in commit: `Reset SuppressEncoding to configured instead of false.` (old) `UnencodedStatementVisitor resets value to previous` (new) makes both tests for `Handlebars_Should_Encode_Chinese` pass.~~

~~Both tests for `Handlebars_Should_Not_Encode_Chinese` still fail (output is Chinese characters, rather than expected &#xxxx;).
I don't know if this should be considered correct behavior, or possibly a separate issue.~~

Looking forward to your comments.

### DIFF
--- a/source/Handlebars.Test/BasicIntegrationTests.cs
+++ b/source/Handlebars.Test/BasicIntegrationTests.cs
@@ -37,6 +37,13 @@ namespace HandlebarsDotNet.Test
     
     public class BasicIntegrationTests
     {
+        private static string HtmlEncodeStringHelper(IHandlebars handlebars, string inputString)
+        {
+            using var stringWriter = new System.IO.StringWriter();
+            handlebars.Configuration.TextEncoder.Encode(inputString, stringWriter);
+            return stringWriter.ToString();
+        }
+
         [Theory]
         [ClassData(typeof(HandlebarsEnvGenerator))]
         public void BasicEnumerableFormatter(IHandlebars handlebars)
@@ -98,7 +105,9 @@ namespace HandlebarsDotNet.Test
                 name = "Handlebars.Net"
             };
             var result = template(data);
-            Assert.Equal("Hello, ('foo' is undefined)!", result);
+
+            var expected = HtmlEncodeStringHelper(handlebars, "Hello, ('foo' is undefined)!");
+            Assert.Equal(expected, result);
         }
         
         [Theory, ClassData(typeof(HandlebarsEnvGenerator))]
@@ -114,7 +123,9 @@ namespace HandlebarsDotNet.Test
                 name = "Handlebars.Net"
             };
             var result = template(data);
-            Assert.Equal("Hello, ('foo' is undefined)!", result);
+
+            var expected = HtmlEncodeStringHelper(handlebars, "Hello, ('foo' is undefined)!");
+            Assert.Equal(expected, result);
         }
         
         [Theory, ClassData(typeof(HandlebarsEnvGenerator))]
@@ -355,7 +366,9 @@ false
                 nestedObject = "Relative dots, yay"
             };
             var result = template(data);
-            Assert.Equal("{ nestedObject = Relative dots, yay }", result);
+
+            var expected = HtmlEncodeStringHelper(handlebars, "{ nestedObject = Relative dots, yay }");
+            Assert.Equal(expected, result);
         }
 
         [Theory, ClassData(typeof(HandlebarsEnvGenerator))]
--- a/source/Handlebars.Test/BasicIntegrationTests.cs
+++ b/source/Handlebars.Test/BasicIntegrationTests.cs
@@ -2013,10 +2013,7 @@ false
 
             var config = new HandlebarsConfiguration
             {
-                Compatibility =
-                {
-                    UseLegacyHandlebarsNetHtmlEncoding = useLegacyHandlebarsNetHtmlEncoding
-                }
+                TextEncoder = useLegacyHandlebarsNetHtmlEncoding ? (ITextEncoder)new HtmlEncoderLegacy() : new HtmlEncoder()
             };
             var actual = Handlebars.Create(config).Compile(template).Invoke(value);
 
@@ -2036,15 +2033,12 @@ false
 
             var config = new HandlebarsConfiguration
             {
-                Compatibility =
-                {
-                    UseLegacyHandlebarsNetHtmlEncoding = !useLegacyHandlebarsNetHtmlEncoding
-                }
+                TextEncoder = !useLegacyHandlebarsNetHtmlEncoding ? (ITextEncoder)new HtmlEncoderLegacy() : new HtmlEncoder()
             };
             var handlebars = Handlebars.Create(config);
+            handlebars.Configuration.TextEncoder = useLegacyHandlebarsNetHtmlEncoding ? (ITextEncoder)new HtmlEncoderLegacy() : new HtmlEncoder();
             var compiledTemplate = handlebars.Compile(template);
-
-            handlebars.Configuration.Compatibility.UseLegacyHandlebarsNetHtmlEncoding = useLegacyHandlebarsNetHtmlEncoding;
+            
             var actual = compiledTemplate(value);
 
             Assert.Equal(expected, actual);
--- a/source/Handlebars.Test/BasicIntegrationTests.cs
+++ b/source/Handlebars.Test/BasicIntegrationTests.cs
@@ -1986,7 +1986,57 @@ false
                 Assert.Equal("42", actual);   
             }
         }
-        
+
+        [Theory]
+        [InlineData(false, "=", "&#x3D;")]
+        [InlineData(true, "=", "=")]
+        public void HtmlEncoderCompatibilityIntegration(bool useLegacyHandlebarsNetHtmlEncoding, string inputChar, string expected)
+        {
+            var template = "{{InputChar}}";
+            var value = new
+            {
+                InputChar = inputChar
+            };
+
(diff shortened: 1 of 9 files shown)


rexm:
What’s the behavior in the JS library for the Chinese characters?

tommysor:
> What’s the behavior in the JS library for the Chinese characters?

I have no idea. Or any idea of how to figure it out.
Closest I see is this page: http://tryhandlebarsjs.com/
But I don't know what settings they are using and if that matches what's relevant for this case.

On http://tryhandlebarsjs.com/ with "expression with raw HTML"
"<" is shown as `&lt;`
While non-ascii characters "öä ñ øå 产品" are shown as entered.

Any insight here @zjklee or @tonysneed ?

tommysor:
I had another look, still cant test if I'm correct.
Looks to me like the JS library does not escape anything except this small list of chars.

https://github.com/handlebars-lang/handlebars.js/blob/master/lib/handlebars/utils.js
```js
const escape = {
  '&': '&amp;',
  '<': '&lt;',
  '>': '&gt;',
  '"': '&quot;',
  "'": '&#x27;',
  '`': '&#x60;',
  '=': '&#x3D;'
};

const badChars = /[&<>"'`=]/g,
  possible = /[&<>"'`=]/;

function escapeChar(chr) {
  return escape[chr];
}

[...]

export function escapeExpression(string) {
  if (typeof string !== 'string') {
    // don't escape SafeStrings, since they're already safe
    if (string && string.toHTML) {
      return string.toHTML();
    } else if (string == null) {
      return '';
    } else if (!string) {
      return string + '';
    }

    // Force a string conversion as this will be done by the append regardless and
    // the regex test will do this transparently behind the scenes, causing issues if
    // an object's to string has escaped characters in it.
    string = '' + string;
  }

  if (!possible.test(string)) {
    return string;
  }
  return string.replace(badChars, escapeChar);
}
```

https://github.com/Handlebars-Net/Handlebars.Net/blob/master/source/Handlebars/IO/HtmlEncoder.cs
```csharp
[...]

private static void EncodeImpl<T>(T text, TextWriter target) where T: IEnumerator<char>
{
	while (text.MoveNext())
	{
		var value = text.Current;
		switch (value)
		{
			case '"':
				target.Write("&quot;");
				break;
			case '&':
				target.Write("&amp;");
				break;
			case '<':
				target.Write("&lt;");
				break;
			case '>':
				target.Write("&gt;");
				break;

			default:
				if (value > 159)
				{
					target.Write("&#");
					target.Write((int)value);
					target.Write(";");
				}
				else target.Write(value);
				break;
		}
	}
}
```

tommysor:
Cleaned up branch. The two commits now address 2 separate issues.

Commit: `UnencodedStatementVisitor resets value to previous`
is a strait forward bugfix, and I believe it should be enough to fix the specific problem in issue #468.

Commit: `HtmlEncoder escape chars based on https://github.com/handlebars-lang/… `
is a danger zone.
Encoding rules copied from JS based on my understanding.
I left out (with comment) 2 chars ( ' and = ) because encoding them breaks existing tests.
Would likely break real world implementations that rely on current behavior.

tonysneed:
@tommysor Thanks for working to fix #468. I'll pull your branch to see if the tests pass in my [repro](https://github.com/tonysneed/handlebars-encoding-test).

@rexm @zjklee I'm looking forward to your review of this PR.

tonysneed:
@tommysor I puled your branch and ran my [tests](https://github.com/tonysneed/handlebars-encoding-test). The tests which set `Handlebars.Configuration.NoEscape = true` all pass.

However, setting `Handlebars.Configuration.NoEscape = false` has no effect. The Chinese characters are not encoded when they should be. Is this expected behavior?

tommysor:
@tonysneed That, I believe, is the $1,000 question for this pull request.

**Under assumption**: Existing encoding rules in Handlebars.Net is correct (https://github.com/Handlebars-Net/Handlebars.Net/blob/master/source/Handlebars/IO/HtmlEncoder.cs).
Then expected result is that the Chinese characters are encoded to &#xxx; when NoEscape = false.
This is what you should get if you cherry pick only the commit `UnencodedStatementVisitor resets value to previous`.
_@tonysneed Your tests will not pass as is. They will pass if you make a change in Class.hbs to double curlys `{{> properties}}` from triple curlys `{{{> properties}}}`._

**Under assumption**: Encoding rules in JavaScript repo is correct (https://github.com/handlebars-lang/handlebars.js/blob/master/lib/handlebars/utils.js).
Then expected result is that the Chinese characters are kept unchanged regardless of NoEscape setting.
This is the current behavior of this branch. Some deviations from js rules would need some cleanup work.

Maintainers (so @rexm and/or @zjklee) should decide the direction for this.
My own suggestion would be that I split the commit `UnencodedStatementVisitor resets value to previous` out into a new pull request. That commit can be released as a patch with very low chance of breaking anything.
While the change to general handling of encoding would be a separate larger change, or possibly not done, according to the maintainers decision.

zjklee:
@tommysor , thanks for taking time for creating the PR.
I'd probably go with a safer option - first merge non-breaking changes. In a separate PR introduce the breaking changes but hide them behind feature toggle. The default behaviour should stay the same in current version but can be changed in the next major release.

source/Handlebars/IO/HtmlEncoder.cs
```diff
@@ -12,6 +12,13 @@ namespace HandlebarsDotNet
     /// </summary>
     public class HtmlEncoder : ITextEncoder
     {
+        private readonly Compatibility _compatibility;
+
+        public HtmlEncoder(Compatibility compatibility)
+        {
+            _compatibility = compatibility;
```
zjklee, line 19: I'd prefer having different encoder implementation.
This approach has at least 2 benefits: 
- easier to deprecate in the future by just removing old implementation
- performance benefit as no `if` would be involved at the time of rendering

zjklee (changes requested):
Can you please also add corresponding documentation to the README?

source/Handlebars/IO/HtmlEncoder.cs
tommysor: Pushed new commit that uses different encoder implementations (replaces bool feature toggle).
Something like this you were thinking of?

I'm delaying change to README until the implementation is confirmed.

source/Handlebars.Test/BasicIntegrationTests.cs
```diff
@@ -37,6 +37,13 @@ public class HandlebarsEnvGenerator : IEnumerable<object[]>
     
     public class BasicIntegrationTests
     {
+        private static string HtmlEncodeStringHelper(IHandlebars handlebars, string inputString)
```
zjklee, line 40: As you've introduced helper method to get `expected` value it might be reasonable to add new entry to `HandlebarsEnvGenerator` to run tests with new encoder as well.

source/Handlebars/IO/HtmlEncoder.cs
```diff
@@ -35,17 +35,18 @@ public void Encode(string text, TextWriter target)
             
             EncodeImpl(text, target);
         }
-        
-        private static void EncodeImpl<T>(T text, TextWriter target) where T: IEnumerator<char>
+
+        private void EncodeImpl<T>(T text, TextWriter target) where T : IEnumerator<char>
```
zjklee, line 39: Why changing it from `static` to instance method?

source/Handlebars/IO/HtmlEncoderLegacy.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Encode<T>(T text, TextWriter target) where T : IEnumerator<char>
+        {
+            if (text == null) return;
+
+            EncodeImpl(text, target);
+        }
+
+        private void EncodeImpl<T>(T text, TextWriter target) where T : IEnumerator<char>
```
zjklee, line 43: Why not `static`?

source/Handlebars/IO/HtmlEncoder.cs
zjklee: Yes, that's what I was thinking about 👍

source/Handlebars.Test/BasicIntegrationTests.cs
tommysor: Added Handlebars with new `HtmlEncoder` to `HandlebarsEnvGenerator`.

source/Handlebars/IO/HtmlEncoder.cs
tommysor: Fixed. No reason, this was left from my first implementation where it did need to be instance method.

source/Handlebars/IO/HtmlEncoderLegacy.cs
tommysor: Fixed. No reason, this was left from my first implementation where it did need to be instance method.

tommysor:
> Can you please also add corresponding documentation to the README?

Added section to README.

zjklee (approved):

zjklee:
Rebase your branch to the latest `master` and it will merge automatically 👍

PR #474, opened 2021-12-01
Issue 469

jamesfarrugia96:
exposed additional helper registration overloads via static methods


zjklee:
Can you add tests to cover new methods?

PR #481, opened 2021-12-20
Support @last by default.

### LINKED ISSUE #480: Supporting `@last` for object iteration by default
For JS implementation, `@last` is supported for object iteration since 2014 (added [here](https://github.com/handlebars-lang/handlebars.js/commit/3a9440f954092558275cd4c05a35ba34bcbfa210)).

For .NET it is disabled by default, and controlled by the `SupportLastInObjectIterations` configuration flag, with the following note:

> If `true` enables support for `@last` in object properties iterations. Not supported in Handlebarsjs.

By leaving the default `false`, the .NET engine is incompatible to the JS engine out-of-the-box.

So please consider changing the default to `true`, or at least add documentation of `SupportLastInObjectIterations`...

Thanks,
Yehoshua

orgads:
Fixes #480

### DIFF
--- a/source/Handlebars.Test/BasicIntegrationTests.cs
+++ b/source/Handlebars.Test/BasicIntegrationTests.cs
@@ -688,7 +688,6 @@ false
         {
             var source = "{{#each enumerateMe}}{{@last}} {{/each}}";
             var template = handlebars.Compile(source);
-            handlebars.Configuration.Compatibility.SupportLastInObjectIterations = true;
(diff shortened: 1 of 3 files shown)


orgads:
How about deprecating it?

yoshigev:
> How about deprecating it?

Sounds good to me. But if not deprecating it, need to add some test setting it to `false`.

zjklee (approved):

PR #482, opened 2021-12-20
Deprecate SupportLastInObjectIterations

orgads:


source/Handlebars/ValueProviders/ObjectIteratorValues.cs
```diff
@@ -7,26 +7,16 @@ namespace HandlebarsDotNet.ValueProviders
...
     {
         private readonly FixedSizeDictionary<ChainSegment, object, ChainSegment.ChainSegmentEqualityComparer> _data;
-        private readonly bool _supportLastInObjectIterations;
-        
+
         private readonly EntryIndex<ChainSegment>[] _wellKnownVariables;
 
         public ObjectIteratorValues(BindingContext bindingContext) : this()
         {
             var configuration = bindingContext.Configuration;
```
zjklee, line 15: Not used anymore

source/Handlebars/ValueProviders/ObjectIteratorValues.cs
```diff
@@ -7,26 +7,16 @@ namespace HandlebarsDotNet.ValueProviders
     public readonly ref struct ObjectIteratorValues
```
zjklee, line 7: After deprecating the flag there's no difference between `ObjectIteratorValues` and `IteratorValues`. `ObjectIteratorValues` should be removed.

zjklee (changes requested):

source/Handlebars/Configuration/Compatibility.cs
```diff
@@ -6,9 +6,10 @@
     public class Compatibility
     {
         /// <summary>
-        /// If <see langword="true"/> enables support for <c>@last</c> in object properties iterations. Not supported in Handlebarsjs.
+        /// If <see langword="true"/> enables support for <c>@last</c> in object properties iterations.
         /// </summary>
-        public bool SupportLastInObjectIterations { get; set; } = false;
+        [System.Obsolete]
```
zjklee, line 11: Add message explaining why it's obsolete and what people should do with it (e.g., remove all usages)

source/Handlebars/ValueProviders/ObjectIteratorValues.cs
orgads: `ObjectIteratorValues` has `Key`, which `IteratorValues` hasn't. Do you want to add it to `IteratorValues`?

orgads: I used inheritance (in a separate commit). If you dislike it, let me know what you do prefer, and I'll update the PR.

zjklee: > `ObjectIteratorValues` has `Key`, which `IteratorValues` hasn't

Oh, I've missed that 😞 

> I used inheritance (in a separate commit). If you dislike it, let me know what you do prefer, and I'll update the PR.

`ObjectIteratorValues` and `IteratorValues` are defined as `readonly ref struct` for performance reasons. I'd prefer having code duplication in this case.

orgads: By the way, [the commit](https://github.com/handlebars-lang/handlebars.js/commit/3a9440f954092558275cd4c05a35ba34bcbfa210) that @yoshigev referenced has this message:
> Provide @key and @last value for all forms of iteration.

So if I read it right, maybe we can just add Key to `IteratorValues` and get rid of the other one?

zjklee: This should work as well. In that case `@Key` should return current index for non-object iterators.

orgads: Done

zjklee:
Removal of `ObjectIteratorValues` introduces breaking change to current v2 API as some external users (e.g., [Handlebars.Net.Extension.NewtonsoftJson](https://github.com/Handlebars-Net/Handlebars.Net.Extension.NewtonsoftJson) and [Handlebars.Net.Extension.Json](https://github.com/Handlebars-Net/Handlebars.Net.Extension.Json)) depends on current behavior.
I'd merge this PR to the new upstream to track [v3 milestone](https://github.com/Handlebars-Net/Handlebars.Net/milestone/2)

zjklee (approved):

orgads:
I can revert the file deletion, and mark it as obsolete. Then you can remove it on v3. WDYT?

zjklee:
> I can revert the file deletion, and mark it as obsolete. Then you can remove it on v3. WDYT?

In that case we'd need to revert addition of `Key` to `IteratorValues` and leave TODO mentioning the intent to do the change.

orgads:
Why? I just re-added the class, and added `[Obsolete]` attribute, so extensions can still use it.

Is it ok?

PR #485, opened 2021-12-26
Add checks for loose closing blocks

### LINKED ISSUE #484: IndexOutOfRangeException during execution of compiled templates
### Describe the bug
Handlebars.net throws  `IndexOutOfRangeException` from  `HandlebarsDotNet.PathStructure.PathResolver.ResolvePath()` on execution  of compiled templates that have loose closing blocks. 

Examples of templates:
`{{#if 0}}test{{/if}}{{/unless}}`
`{{#if 1}}{{#unless 0}}test{{/if}}{{/unless}}{{/if}}`

### Expected behavior:
Handlebars.net compiler should catch this type of template errors during parsing and building expression trees and throw `HandlebarsCompilerException` with a meaningful error message.

### Test to reproduce
```csharp
[Fact]
public void TestLooseClosingBlockExpressionException()
{ 
    Assert.Throws<HandlebarsCompilerException>(() =>
    {
        Handlebars.Compile("{{#if 0}}test{{/if}}{{/unless}}")(new { });
    });
}
```

abraham-fox:
Check loose closing blocks on compile time - fixes #484.

### DIFF
--- a/source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
+++ b/source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
@@ -10,7 +10,7 @@ namespace HandlebarsDotNet.Compiler
         private static readonly HashSet<string> ConditionHelpers = new HashSet<string>(StringComparer.OrdinalIgnoreCase){ "#if", "#unless", "^if", "^unless" };
         private static readonly HashSet<string> IteratorHelpers = new HashSet<string>(StringComparer.OrdinalIgnoreCase){ "#each", "^each" };
 
-        public static BlockAccumulatorContext Create(Expression item, ICompiledHandlebarsConfiguration configuration)
+        public static BlockAccumulatorContext Create(Expression item, Expression parentItem, ICompiledHandlebarsConfiguration configuration)
         {
             BlockAccumulatorContext context = null;
             if (IsConditionalBlock(item))
@@ -28,6 +28,10 @@ namespace HandlebarsDotNet.Compiler
             else if (IsBlockHelper(item, configuration))
             {
                 context = new BlockHelperAccumulatorContext(item);
+            }
+            else if (IsLooseClosingElement(item, parentItem, out var looseBlockName))
+            {
+                throw new HandlebarsCompilerException($"Loose closing block '{looseBlockName}' was found");
             }
 
             return context;
@@ -74,6 +78,54 @@ namespace HandlebarsDotNet.Compiler
                 
                 default:
                     return false;
+            }
+        }
+
+        private static bool IsLooseClosingElement(Expression item, Expression parentItem, out string looseBlockName)
+        {
+            looseBlockName = null;
+
+            var itemBlockName = GetBlockName(item);
+
+            if (itemBlockName == null) return false;
+
+            var parentBlockName = GetBlockName(parentItem);
+
+            if (!itemBlockName.StartsWith("/")) return false;
+
+            if (parentBlockName == null || IsClosingBlockNotMatchParentBlock(itemBlockName, parentBlockName))
+            {
+                looseBlockName = itemBlockName;
+
+                return true;
+            }
+
+            return false;
+        }
+
+        private static bool IsClosingBlockNotMatchParentBlock(string itemBlockName, string parentBlockName)
+        {
+            if (itemBlockName == null) throw new ArgumentNullException(nameof(itemBlockName));
+            if (parentBlockName == null) throw new ArgumentNullException(nameof(parentBlockName));
+
+            if (!parentBlockName.StartsWith("#") || parentBlockName.StartsWith("#>") || parentBlockName.StartsWith("#*")) return false;
+
+            return parentBlockName.Substring(1) != itemBlockName.Substring(1);
+        }
+
+        private static string GetBlockName(Expression item)
+        {
+            item = UnwrapStatement(item);
+            switch( item )
+            {
+                case PathExpression pathExpression:
+                    return pathExpression.Path;
+
+                case HelperExpression helperExpression:
+                    return helperExpression.HelperName;
+
+                default:
+                    return null;
             }
         }
 
--- a/source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
+++ b/source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq.Expressions;
 using HandlebarsDotNet.PathStructure;
+using HandlebarsDotNet.StringUtils;
 
 namespace HandlebarsDotNet.Compiler
 {
@@ -68,17 +69,12 @@ namespace HandlebarsDotNet.Compiler
         private static bool IsPartialBlock (Expression item)
         {
             item = UnwrapStatement (item);
-            switch (item)
+            return item switch
             {
-                case PathExpression expression:
-                    return expression.Path.StartsWith("#>");
-                
-                case HelperExpression helperExpression:
-                    return helperExpression.HelperName.StartsWith("#>");
-                
-                default:
-                    return false;
-            }
+                PathExpression expression => expression.Path.StartsWith("#>"),
+                HelperExpression helperExpression => helperExpression.HelperName.StartsWith("#>"),
+                _ => false,
+            };
         }
 
         private static bool IsLooseClosingElement(Expression item, Expression parentItem, out string looseBlockName)
@@ -110,23 +106,18 @@ namespace HandlebarsDotNet.Compiler
 
             if (!parentBlockName.StartsWith("#") || parentBlockName.StartsWith("#>") || parentBlockName.StartsWith("#*")) return false;
 
-            return parentBlockName.Substring(1) != itemBlockName.Substring(1);
+            return new Substring(parentBlockName, 1) != new Substring(itemBlockName, 1);
         }
 
         private static string GetBlockName(Expression item)
         {
             item = UnwrapStatement(item);
-            switch( item )
+            return item switch
             {
-                case PathExpression pathExpression:
-                    return pathExpression.Path;
-
-                case HelperExpression helperExpression:
-                    return helperExpression.HelperName;
-
-                default:
-                    return null;
-            }
+                PathExpression pathExpression => pathExpression.Path,
+                HelperExpression helperExpression => helperExpression.HelperName,
+                _ => null,
+            };
         }
 
         protected static Expression UnwrapStatement(Expression item)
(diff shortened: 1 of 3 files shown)


source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
```diff
@@ -77,6 +81,54 @@ private static bool IsPartialBlock (Expression item)
...
+        }
+
+        private static bool IsClosingBlockNotMatchParentBlock(string itemBlockName, string parentBlockName)
+        {
+            if (itemBlockName == null) throw new ArgumentNullException(nameof(itemBlockName));
+            if (parentBlockName == null) throw new ArgumentNullException(nameof(parentBlockName));
+
+            if (!parentBlockName.StartsWith("#") || parentBlockName.StartsWith("#>") || parentBlockName.StartsWith("#*")) return false;
+
+            return parentBlockName.Substring(1) != itemBlockName.Substring(1);
```
zjklee, line 113: Why can't you just do `parentBlockName[0] != itemBlockName[0]`? `Substring` will produce new string allocation.

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
```diff
@@ -77,6 +81,54 @@ private static bool IsPartialBlock (Expression item)
...
+            {
+                case PathExpression pathExpression:
+                    return pathExpression.Path;
+
+                case HelperExpression helperExpression:
+                    return helperExpression.HelperName;
+
+                default:
+                    return null;
+            }
```
zjklee, line 129: nit: Use [switch expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression#:~:text=switch%20expression%20%28C%23%20reference%29%201%20Case%20guards.%20A,an%20exception.%20...%203%20C%23%20language%20specification.%20)

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
```diff
@@ -77,6 +81,54 @@ private static bool IsPartialBlock (Expression item)
...
+                return true;
+            }
+
+            return false;
+        }
+
+        private static bool IsClosingBlockNotMatchParentBlock(string itemBlockName, string parentBlockName)
+        {
+            if (itemBlockName == null) throw new ArgumentNullException(nameof(itemBlockName));
+            if (parentBlockName == null) throw new ArgumentNullException(nameof(parentBlockName));
```
zjklee, line 109: nit: Shouldn't it be `string.IsNullOrEmpty` check?

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
abraham-fox: I'm not sure that I've understood you correctly because `parentBlockName[0] != itemBlockName[0]` will just compare "#" with "/" and it will cause a false positive alarm for any correct statement like `{{#if}}test{{/if}}`. Here the code compares if an open element matches with a closing element (e.g. `#if` with `/unless`).

abraham-fox: Yes, I'm aware of string allocations and the best I can do (as far as I know) is to use `MemoryExtensions.AsSpan` method but it's available only from .NET Standard 2.1. So should I use preprocessor directive `#if netstandard2_1 ...` for this case or do you know a better solution?

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
abraham-fox: Yeah, as I did it initially, but then I noticed another code block with this style and formated my code similarly because of
> Contributing
> ...
> Follow the established code format

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
abraham-fox: I used this check to ensure that this function won't be used incorrectly in other places (just in case). It checks only null because `GetBlockName()` returns null as default (when a statement is nor PathExpression nor HelperExpression).

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
zjklee: > I'm not sure that I've understood you correctly because parentBlockName[0] != itemBlockName[0] will just compare "#" with "/"

My bad, I've interpreted the code in the wrong way. 

> Yes, I'm aware of string allocations and the best I can do (as far as I know) is to use `MemoryExtensions.AsSpan` method but it's available only from .NET Standard 2.1. So should I use preprocessor directive `#if netstandard2_1 ...` for this case or do you know a better solution?

You may use `Substring` struct:
```suggestion
            return new Substring(parentBlockName, 1) != new Substring(itemBlockName, 1);
```
Based on benchmarking it behaves equivalently from execution time but better in terms of allocations:

|                 Method |  N |      Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------- |--- |----------:|----------:|----------:|-------:|------:|------:|----------:|
|            EqualString |  2 |  4.741 ns | 0.0186 ns | 0.0156 ns |      - |     - |     - |         - |
|            EqualString |  5 | 29.061 ns | 0.2653 ns | 0.2482 ns | 0.0204 |     - |     - |      64 B |
|            EqualString | 10 | 27.998 ns | 0.1122 ns | 0.0937 ns | 0.0255 |     - |     - |      80 B |
|                        |    |           |           |           |        |       |       |           |
|         EqualSubstring |  2 | 22.192 ns | 0.0824 ns | 0.0730 ns |      - |     - |     - |         - |
|         EqualSubstring |  5 | 20.849 ns | 0.0441 ns | 0.0413 ns |      - |     - |     - |         - |
|         EqualSubstring | 10 | 20.914 ns | 0.0739 ns | 0.0691 ns |      - |     - |     - |         - |
|                        |    |           |           |           |        |       |       |           |
|    NotEqualStartString |  2 | 26.196 ns | 0.1376 ns | 0.1220 ns | 0.0204 |     - |     - |      64 B |
|    NotEqualStartString |  5 | 26.828 ns | 0.4037 ns | 0.3371 ns | 0.0204 |     - |     - |      64 B |
|    NotEqualStartString | 10 | 29.938 ns | 0.2559 ns | 0.2393 ns | 0.0306 |     - |     - |      96 B |
|                        |    |           |           |           |        |       |       |           |
| NotEqualStartSubstring |  2 | 23.794 ns | 0.0685 ns | 0.0641 ns |      - |     - |     - |         - |
| NotEqualStartSubstring |  5 | 25.827 ns | 0.1304 ns | 0.1089 ns |      - |     - |     - |         - |
| NotEqualStartSubstring | 10 | 25.739 ns | 0.0882 ns | 0.0737 ns |      - |     - |     - |         - |
|                        |    |           |           |           |        |       |       |           |
|      NotEqualEndString |  2 | 28.473 ns | 0.2956 ns | 0.2765 ns | 0.0204 |     - |     - |      64 B |
|      NotEqualEndString |  5 | 28.210 ns | 1.7463 ns | 1.5481 ns | 0.0204 |     - |     - |      64 B |
|      NotEqualEndString | 10 | 30.028 ns | 0.3705 ns | 0.3466 ns | 0.0306 |     - |     - |      96 B |
|                        |    |           |           |           |        |       |       |           |
|   NotEqualEndSubstring |  2 | 26.405 ns | 0.0686 ns | 0.0608 ns |      - |     - |     - |         - |
|   NotEqualEndSubstring |  5 | 25.046 ns | 0.0434 ns | 0.0406 ns |      - |     - |     - |         - |
|   NotEqualEndSubstring | 10 | 29.412 ns | 0.1160 ns | 0.1085 ns |      - |     - |     - |         - |

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
zjklee: I'd prefer to utilize new features when applicable 😄

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
```diff
@@ -29,6 +29,10 @@ public static BlockAccumulatorContext Create(Expression item, ICompiledHandlebar
             {
                 context = new BlockHelperAccumulatorContext(item);
             }
+            else if (IsLooseClosingElement(item, parentItem, out var looseBlockName))
+            {
+                throw new HandlebarsCompilerException($"Loose closing block '{looseBlockName}' was found");
```
rexm, line 34: Dropping this in from my mobile to start a conversation; I can follow up with more specific suggestions: I am not sure the exception message or phrase “loose closing block” is clear or actionable enough by itself for many devs. They would need to look that up to understand it. Can we make it more clear what the problem is?

abraham-fox: I see your point and I had worries about the wording. What do you think about "An open element was missed for the closing element '/unless'"?

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
abraham-fox: Thank you, I didn't know about this useful utility function. Updated!

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
abraham-fox: Me too! Done.

source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/BlockAccumulatorContext.cs
rexm: Now that we're adding more specific compile-time messages, I'm really wishing I'd added the ability to keep line number and source context :) 

"A closing element {{/unless}} was found without a matching open element"?

zjklee: > Now that we're adding more specific compile-time messages, I'm really wishing I'd added the ability to keep line number and source context :)

@rexm , I've added this some time ago but unfortunately it's not propagated to every corner of the compiler 🙁

rexm: Oh awesome! I’ll need to look at how that works :)

abraham-fox: The message was updated. Also, I've made renaming for functions and vars to get rid of some ambiguity.

zjklee (approved):

Issue #500, opened 2022-02-08
Last letter cut off when using keys set in context since v2.0.3?


zjklee:
### Discussed in https://github.com/Handlebars-Net/Handlebars.Net/discussions/499

<div type='discussions-op-text'>

<sup>Originally posted by **cletom** February  4, 2022</sup>
Hi,

I always used to get general info via Handlebars from context.

Therefore, I defined the values as follows:

```
var context = ImmutableDictionary<string, object>.Empty
                .Add("Name", "abcd");
```

and I was able to receive the value I set using

`{{.Name}}`

which worked fine until v 2.0.2.

When doing the same thing with v2.0.3 or later I always get an Error telling me `Nam is undefined` so it seems like somehow the last letter gets cut off since v2.0.3. If I add any sign, e.g. if I use `{{.Name1}}` i receive the value I set for "Name". I already debuged into my code and the context contains the correct key/value pairs.

Does anyone know what I have to do to make this work again or is there a bug in the library since v2.0.3?

Thanks</div>

PR #504, opened 2022-02-23
Last letter cut off when the string starts with the separator

periklis92:
Split Enumerator strings would have the wrong length if the string started with the separator.
This refers to issue #500 and this fix seems to have solved it.

### DIFF
--- a/source/Handlebars.Test/IssueTests.cs
+++ b/source/Handlebars.Test/IssueTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Dynamic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


rexm:
Thanks for the contribution! Since this bug slipped in without failing any unit tests, can you please add at least 1 so we can guarantee it never happens again?

periklis92:
Sure thing!

periklis92:
So, I was writing some tests and I noticed something, and I am not sure if it is intended or a side-effect of my change.
When I split a string, with no StringSplitOptions, that starts with the separator like "/a/c/d/e" the first element of the Enumerator is an empty string (""), and the rest are "a", "c", "d", "e" as expected.

rexm:
If you revert your change does it still happen?

periklis92:
You are right. It does.

zjklee:
@periklis92 , the problem looks to be in Github actions. See https://github.com/Handlebars-Net/Handlebars.Net/pull/506 for more details. Once the PR is merged you may rebase your branch on top of the latest.

zjklee:
@periklis92 , can you please add a test that covers scenario from the issue?

periklis92:
The build seems to work now. I'll fix the code smell and clean the commits before the final push. 
@zjklee , when you say add a test, you mean regarding the netfx451 on windows 2022 issue, or something else?

zjklee:
@periklis92 , I mean `{{.Name}} ` scenario from the issue.

zjklee (approved):

PR #527, opened 2022-10-17
Add one More Compile Method Overload

HaruTzuki:
I am working in a project and I thought this will be a great method.

### DIFF
--- a/source/Handlebars.Test/BasicIntegrationTests.cs
+++ b/source/Handlebars.Test/BasicIntegrationTests.cs
@@ -2085,6 +2085,20 @@ false
             Assert.Equal("abc", nestedObjectsHelperResult);
         }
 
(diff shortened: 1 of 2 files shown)


rexm:
I'm curious what is the use case for this?

HaruTzuki:
In order to compile a text with handlerbars you need first to compile the text to take the template and after that you need to pass the object. I decided it will be great this to happens in one line if you need to use it in a lamda expression e.g.

`public string GetText(string source, object handles)=>Handlebars.Compile(source, handles); // something like that.`

HaruTzuki:
Ok then... Delete the PR.

rexm:
Happy to reopen if anyone feels strongly enough about this to address the ergonomics issue.

Issue #524, opened 2022-08-11
Template partial keeps getting html escaped


BartVanBeurden:
I have a layout partial

partials/layout.hbs
```
<html>
<body>
    <div class="content">
        {{> content}}
    </div>
</body>
</html>
```

and a message handlebars that uses this layout

message.hbs
```
{{#> layout}}

    {{#*inline "content"}}
        <b>{{ Data.Message }}</b>
    {{/inline}}

{{/layout}}
```

With the default configuration all my html tags are getting escaped (both in the layout.hbs as well as in the inline content hbs). Why? How do I fix this?
I don't believe this is normal behaviour for handlebars.js

The only way to render this properly is to pass in `NoEscape` in the configuration, but this has the unwanted side-effect that my {{ Data.Message }} does not get sanitized and can also contain html.

zjklee:
Hello @BartVanBeurden 
Looks like a bug to me. Can you please provide a test covering your scenario?

Nisden:
@zjklee I have created a PR with an unit test demonstrating the issue.

Nisden:
@zjklee I implemented a fix, but I am not 100% certain its correct, so please take a look.

Ringconfined688:
@Nisden https://rapacity30.github.io/ecstatic/

PR #538, opened 2022-12-22
Rendering inline blocks are incorrectly escaped

Nisden:
This pull request demonstrates the the issue experienced in #524


rexm:
Seems OK to me. May be a minute before it's published through NuGet, @zjklee will need do to that as I'm no longer set up for it.

Nisden:
Thank you!

PR #562, opened 2023-12-08
Fix LiteralConverter to support long

StefH:


StefH:
@oformaniuk / @rexm SonarCloud is fixed.

Can you please review this PR?

source/Handlebars/Arguments.cs
```diff
@@ -58,7 +58,7 @@ public Arguments(object arg1) : this()
             Length = 1;
         }
         
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        // [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
oformaniuk, line 61: Why do you need to comment out the attribute?

oformaniuk (approved):

source/Handlebars/Arguments.cs
StefH: Sorry, this was a debugging leftover.

PR #567, opened 2024-02-14
Allow slashes properly within escape blocks

### LINKED ISSUE #566: Consecutive slashes and more than 1 slash within square brackets does not escape properly
### Describe the bug

We have a use case where our JSON data contains a field like this:

```
{
  "https://some-uri.com/foo/bar_baz": {
    "id": 42
  }
}
```

And we want to parse out the 42 like so:
```
{{[https://some-uri.com/foo/bar_baz].id}}
```

### Expected behavior:

According to the handlebars spec and various other implementations, this should yield `42`. However, we instead see:
```
Unhandled exception. System.IndexOutOfRangeException: Index was outside the bounds of the array.
```

After some testing we figured out it is caused by `//` in the template string. We then tried removing just `//`:

```
{
  "https:some-uri.com/foo/bar_baz": {
    "id": 42
  }
}

{{[https:some-uri.com/foo/bar_baz].id}}
```

While the above doesn't throw an exception, it fails to resolve at all. (returns empty)

Some more debugging, we found that it is caused by the forward slashes, as the following does properly return 42 (keeping at most one `/`, not multiple):

```
{
  "https:some-uri.com/foobar_baz": {
    "id": 42
  }
}

{{[https:some-uri.com/foobar_baz].id}}
```

We expect that the original case properly returns 42, as `[ ]` should properly escape the whole URI. It seems like the current HandleBars.Net implementation special-cases escaping of `/` only for a single `/` within square brackets, and multiple consecutive `/` gives an unexpected exception.

### Test to reproduce
```csharp
using HandlebarsDotNet;
using HandlebarsDotNet.Extension.NewtonsoftJson;
using Newtonsoft.Json.Linq;

[Fact]
public void Descriptive_Test_Name_Here()
{
    var source = "{{[https://some-uri.com/foo/bar_baz].id}}";
    var dataJson = "{\"https://some-uri.com/foo/bar_baz\":{\"id\":42}}";

    handlebars.Configuration.UseNewtonsoftJson();

    var template = handlebars.Compile(source);
    var actual = template(JObject.Parse(dataJson));

    Assert.Equal(42, actual);
}
```

### Other related info

---

### After submitting the issue
Please consider contributing to the project by submitting a PR with a fix to the issue. 
This would help to solve your problem in a shorter time as well as help other users of the project.

In case you do not know where to start - feel free to ask for help in the issue thread.

Building an active community is essential for any project survival as time of maintainers is limited.

Hoeksema:
closes #566 

The path parsing currently doesn't work properly when there are embedded slashes within an ignore block.

This PR fixes this issue:
- No more exceptions thrown when using `//` within an escaped block
- Allowing multiple `/` to occur within an escape block without breakage

Before, the individual segments between slashes in addition to the entire escaped block were returned by PathInfo. Now, it returns just the latter, which is correct. All existing unit tests still pass and new tests were added to exercise the failing cases in #566.

### DIFF
--- a/source/Handlebars.Test/PathInfoTests.cs
+++ b/source/Handlebars.Test/PathInfoTests.cs
@@ -50,6 +50,8 @@ namespace HandlebarsDotNet.Test
         [InlineData("a/[b.c].[b/c]/d", new [] {"a", "[b.c].[b/c]", "d"})]
         [InlineData("a/[b/c]/d", new [] {"a", "[b/c]", "d"})]
         [InlineData("a/[b.c/d]/e", new [] {"a", "[b.c/d]", "e"})]
+        [InlineData("a/[b.c/d/e/f]/e", new [] {"a", "[b.c/d/e/f]", "e"})]
+        [InlineData("a/[a//b/c.d/e]/e", new [] {"a", "[a//b/c.d/e]", "e"})]
         public void SlashPath(string input, string[] expected)
         {
             var pathInfo = PathInfo.Parse(input);
--- a/source/Handlebars.Test/PathInfoTests.cs
+++ b/source/Handlebars.Test/PathInfoTests.cs
@@ -52,6 +52,7 @@ namespace HandlebarsDotNet.Test
         [InlineData("a/[b.c/d]/e", new [] {"a", "[b.c/d]", "e"})]
         [InlineData("a/[b.c/d/e/f]/e", new [] {"a", "[b.c/d/e/f]", "e"})]
         [InlineData("a/[a//b/c.d/e]/e", new [] {"a", "[a//b/c.d/e]", "e"})]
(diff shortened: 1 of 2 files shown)


Hoeksema:
Hey @oformaniuk @StefH, hope you're having a great week! Any chance you could review this?

We're looking to keep our code base unforked from origin as much as possible, also so we can keep sending pull requests for issues we find as we productionize this in our system, and need this fix in particular for a critical use case of ours.

oformaniuk:
@Hoeksema, unfortunately, I was not following the repo closely lately. If you (and your team) is interested in contributing into the project further I'd be happy to include you as contributors.

Hoeksema:
> @Hoeksema, unfortunately, I was not following the repo closely lately. If you (and your team) is interested in contributing into the project further I'd be happy to include you as contributors.

Thank you for merging the PRs! Feel free to add me as contributor if that feels appropriate. We might need to make some more changes in a quarter or two once we use it more broadly.

PR #570, opened 2024-02-16
Add EmbedUntrackedSources

lahma:
* added `EmbedUntrackedSources` to remove one warning, see https://nuget.info/packages/Handlebars.Net/2.1.4

I would also suggest changing to use newer GH Actions images for building so that other warnings would go away (old SDK in use). Maybe another modernization step could be removing old unsupported full framework targets and only support oldest supported `net462`. Adding `net6.0` target would allow one target without dependency on `Microsoft.Csharp`.


StefH:
@lahma 
Thank you.
But actually the "license expression" is already fixed by me (https://github.com/Handlebars-Net/Handlebars.Net/pull/564)

About EmbedUntrackedSources  --> that's ok.

And thanks about the hint for "newer GH Actions images", I'll take a look.

Issue #573, opened 2024-03-31
Improve dependency graph by adding TFM


thompson-tomo:
**Is your feature request related to a problem? Please describe.**
I want to minimise dependencies in my project by utilising framework dependencies wherever possible

**Describe the solution you'd like**
Additional TFM'S are added so that The Microsoft.CSharp package becomes conditional and is only added for the older frameworks

**Describe alternatives you've considered**
Accept the additional dependency

**Additional context**
n/a

PR #577, opened 2024-04-01
Resolve merge conflicts + cleanup csproj

thompson-tomo:
Will handle #573


StefH:
@thompson-tomo and @oformaniuk 
I think it's more clear that this PR is just closed [because this PR now contains more changed : **87** files changed].
Better to create a new PR which only adds the new frameworks.

Side note: do we actually want to add .NET 5 because that one is out of support for some time? And instead just add .NET 6 and .NET 8 because .NET 7 is also almost out of official support?

Issue #415, opened 2021-01-20
Any porting available for .net core 5.0


Nagendracv:
Looking forward to  .net core support for this library

zjklee:
Hello @Nagendracv 
Library supports netstandard 2.0 and 2.1 so you should be able to use it with net5 without any problems.
PR with net5 target is welcome 😀

huysentruitw:
I'm using this successfully in a .NET 5 project, so not sure what the question here is referring to. Can you elaborate?

jonfast565:
Can also use with .NET 5.

PR #578, opened 2024-04-01
Update of TFM & dependency optimisation

thompson-tomo:
By adjusting the TFM'S we have been able to produce a package with no dependencies on the latest frameworks hence an optimised dependency graph.

The following frameworks have been added:
- Net 6

The following frameworks even though requested was not added:
- Net 5

The following frameworks were removed:
- Net 4.5.2
- Net 4.6

Closes: #573
Closes: #415

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   push:
     branches: [ master ]
- 
+
 jobs:
   build:
     name: Build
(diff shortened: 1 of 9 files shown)


oformaniuk (changes requested):
Consider dropping `net5` as it is not supported anymore and adding `net8`.

thompson-tomo:
I don't see a benefit in adding dotnet 8 as the dotnet 6 compilation will be used. Will remove net 5.

Thoughts on also dropping net 452, net 46 as it is the same as th3 net 451 compilation.

oformaniuk (approved):

oformaniuk:
> Thoughts on also dropping net 452, net 46 as it is the same as the net 451 compilation.

Yes, this is a good idea to optimize monikers 👍

thompson-tomo:
@oformaniuk i have just gone & dropped those 2 TFM's from the package.

oformaniuk:
@thompson-tomo , CI pipeline is now broken: https://github.com/Handlebars-Net/Handlebars.Net/actions/runs/8524444266/job/23349047350

thompson-tomo:
@oformaniuk should be Fixed in #576

Issue #285, opened 2019-02-18
How to treat 0 as truthy


sellomkantjwa:
I tried using 

```
testing Var: 
{{#if var includeZero=true}}
    var is set to {{var}}
{{/if}}
```
but I get 
```
Unhandled Exception: System.InvalidOperationException: Sequence contains more than one element
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source)
   at HandlebarsDotNet.Compiler.ConditionalBlockAccumulatorContext..ctor(Expression startingNode)
   at HandlebarsDotNet.Compiler.BlockAccumulatorContext.Create(Expression item, HandlebarsConfiguration configuration)
   at HandlebarsDotNet.Compiler.BlockAccumulator.ConvertTokens(IEnumerable`1 sequence)+MoveNext()
   at System.Collections.Generic.List`1.AddEnumerable(IEnumerable`1 enumerable)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at HandlebarsDotNet.Compiler.ExpressionBuilder.ConvertTokensToExpressions(IEnumerable`1 tokens)
   at HandlebarsDotNet.Compiler.HandlebarsCompiler.Compile(TextReader source)
   at HandlebarsDotNet.Handlebars.HandlebarsEnvironment.Compile(String template)

```

rexm:
Handlebars doesn’t support arbitrary statements, only helpers.

sellomkantjwa:
Cool thanks, 
Turns out this syntax applies to the JS parser:
https://github.com/wycats/handlebars.js/issues/608

tyriker:
@rexm do you mean `includeZero=true` as being an arbitrary statement? The `includeZero` option has been a part of the `#if` helper since roughly 2014. https://handlebarsjs.com/guide/builtin-helpers.html#if

Should Handlebars.NET support this option? I expected it to work and saw the same result.

I believe the exception comes from here https://github.com/Handlebars-Net/Handlebars.Net/blob/9cf16727d3acc2efa6ab180e8f8a90c2e2b301ee/source/Handlebars/Compiler/Lexer/Converter/BlockAccumulators/ConditionalBlockAccumulatorContext.cs#L33 -- which is just the reason for the Exception, not the underlying problem/solution.

I have forked the repo to take a closer look, but can't promise anything and may opt for a workaround as it relates to my current needs.

rexm:
[...]

PR #600, opened 2025-03-01
Added support for includeZero parameter for conditional blocks

AidenFuller:
Added support for includeZero hash parameter for inbuilt conditional blocks as mentioned by #285.


AidenFuller:
Is a pipeline update required here to resolve the failing checks? I'm not much of a CI wiz, but those errors don't look related to this change specifically

PR #604, opened 2025-08-21
WeakReference garbage collection

abasau-incomm:
While updating Handlebars.Net from 1.x to 2.1.6, I have noticed that GC Handler count increases significantly with every compiled template. GC Handlers are associated with WeakReference objects which were introduced in https://github.com/Handlebars-Net/Handlebars.Net/pull/456.

I discovered a couple of issues while investigating the GC Handlers increase.

My app complies templates using the static method `Handlebars.Compile()`.

### Before the Update
image

### After the Update
image

### DIFF
--- a/source/Handlebars/Collections/DictionarySlim.cs
+++ b/source/Handlebars/Collections/DictionarySlim.cs
@@ -183,7 +183,10 @@ namespace HandlebarsDotNet.Collections
                 (uint)i < (uint)entries.Length; i = entries[i].Next)
             {
                 if (_comparer.Equals(key, entries[i].Key))
+                {
                     entries[i].Value = value;
+                    return;
+                }
                 if (collisionCount == entries.Length)
                 {
                     // The chain of entries forms a loop; which means a concurrent update has happened.
--- a/source/Handlebars/Collections/WeakCollection.cs
+++ b/source/Handlebars/Collections/WeakCollection.cs
@@ -13,6 +13,8 @@ namespace HandlebarsDotNet.Collections
         
         public void Add(T value)
         {
+            // Need a way to reset _firstAvailableIndex periodically
+
             for (var index = _firstAvailableIndex; index < _store.Count; index++)
             {
                 if (_store[index] == null)
--- a/source/Handlebars/Collections/WeakCollection.cs
+++ b/source/Handlebars/Collections/WeakCollection.cs
@@ -13,8 +13,6 @@ namespace HandlebarsDotNet.Collections
         
         public void Add(T value)
         {
-            // Need a way to reset _firstAvailableIndex periodically
-
             for (var index = _firstAvailableIndex; index < _store.Count; index++)
             {
                 if (_store[index] == null)
@@ -31,9 +29,13 @@ namespace HandlebarsDotNet.Collections
                     return;
                 }
             }
-            
+
             _store.Add(new WeakReference<T>(value));
-            _firstAvailableIndex = _store.Count;
+            // Reset to 0 so the next Add() scans the full list and can reclaim any slots
+            // that became dead (GC-collected) since the last full pass. Without this reset,
+            // _firstAvailableIndex stays pinned at _store.Count after every append, dead
+            // slots are never reused, and the WeakReference/GC-handle count grows without bound.
+            _firstAvailableIndex = 0;
         }
         
         public void Remove(T value)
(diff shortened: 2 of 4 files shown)


source/Handlebars/Collections/DictionarySlim.cs
```diff
@@ -183,7 +183,10 @@ public void AddOrReplace(in TKey key, in TValue value)
                 (uint)i < (uint)entries.Length; i = entries[i].Next)
             {
                 if (_comparer.Equals(key, entries[i].Key))
+                {
                     entries[i].Value = value;
+                    return;
+                }
```
abasau-incomm, line 189: `AddOrReplace()` method always adds a key-value pair even if the key was found and replaced in the dictionary. This issue can be observed by inspecting the configuration object of Handlebars with a shared environment (`Handlebars.CreateSharedEnvironment(configuration)`) after multiple compilations.

image
image

source/Handlebars/Pools/BindingContext.Pool.cs
```diff
@@ -46,6 +46,8 @@ internal struct BindingContextPolicy : IInternalObjectPoolPolicy<BindingContext>
 
                 public bool Return(BindingContext item)
                 {
+                    item.Configuration = null;
+
                     item.Root = null;
```
abasau-incomm, line 51: `BindingContextPool` may temporary prevent garbage collection of some configuration objects because the `Configuration` property isn't cleared when BindingContext is being returned to the pool. It's unlikely that this issue will cause actual problems because the pool will eventually cycle through all BindingContext objects and it will release the configuration object by assigning a new configuration object on:
https://github.com/Handlebars-Net/Handlebars.Net/blob/9cf16727d3acc2efa6ab180e8f8a90c2e2b301ee/source/Handlebars/Pools/BindingContext.Pool.cs#L32

But it makes sense to clear Configuration property when returning BindingContext to the pool because this property will be set to a new value in the `CreateContext()` method. It's better to clear it explicitly as not to prevent GC.

image

rexm:
Thanks for working on this, I’m traveling but will keep an eye on it!

source/Handlebars/Collections/WeakCollection.cs
```diff
@@ -13,6 +13,8 @@ public class WeakCollection<T> : IEnumerable<T> where T : class
         
         public void Add(T value)
         {
+            // Need a way to reset _firstAvailableIndex periodically
+
             for (var index = _firstAvailableIndex; index < _store.Count; index++)
```
abasau-incomm, line 16: And this spot is the main source of constant growth of GC Handlers and WeakReferences. **I don't have a good idea about how to fix it and I'd appreciate some guidance.**

This issue can only be observed if a static method `Handlebars.Compile(source)` is used, or if a Handlebars object is stored in a static variable (e.g. `private static IHandlebars _handlebars = Handlebars.Create();`), or `HandlebarsConfiguration` object is stored in a static variable.

Every time when a template is compiled, new instances of `FormatterProvider` and `ObjectDescriptorFactory` are created on:
https://github.com/Handlebars-Net/Handlebars.Net/blob/9cf16727d3acc2efa6ab180e8f8a90c2e2b301ee/source/Handlebars/HandlebarsEnvironment.cs#L132-L133

Those new instances are then subscribed to the original FormatterProvider and ObjectDescriptorFactory objects from the main configuration object.

WeakCollection has a mechanism to detect dead references and re-use WeakReference objects to store new values. It's done by resetting `_firstAvailableIndex` in `Remove()` and `GetEnumerator()`. But, if those two methods are never called, `List<WeakReference<T>` grows continuously.

image

image

`ObservableIndex` and `ObservableList` that utilize `WeakCollection` have `Subscribe()` methods which return a disposable container which calls `Remove()` when disposed.

https://github.com/Handlebars-Net/Handlebars.Net/blob/9cf16727d3acc2efa6ab180e8f8a90c2e2b301ee/source/Handlebars/Collections/ObservableIndex.cs#L34-L52

But none of the methods that call `Subscribe()` save those disposable containers and therefore don't dispose them.

So maybe the fix should be to keep track of disposable containers in the methods that call `Subscribe()` and dispose them in finalizers?

abasau-incomm:
> Thanks for working on this, I’m traveling but will keep an eye on it!

@rexm thanks! I am still adding details to the PR and will hit "Ready for review" once done.

abasau-incomm:
@rexm thank you so much for taking over the PR! I wouldn't have enough confidence and knowledge to finish the fix beside trivial changes.

PR #606, opened 2025-10-17
Allow empty partial-block

TheConstructor:
I authored a partial-template with optional block-includes and happened to write `{{#>mail}}{{/mail}}` for the test-mail, as I did not need to adjust the optionals. This led to an exception similar to the one below:

```
System.InvalidOperationException
Sequence contains no elements
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at HandlebarsDotNet.Compiler.PartialBlockAccumulatorContext.GetAccumulatedBlock() in ..\Handlebars.Net\source\Handlebars\Compiler\Lexer\Converter\BlockAccumulators\PartialBlockAccumulatorContext.cs:line 27
   at HandlebarsDotNet.Compiler.BlockAccumulator.AccumulateBlock(Expression parentItem, IEnumerator`1 enumerator, BlockAccumulatorContext context) in ..\Handlebars.Net\source\Handlebars\Compiler\Lexer\Converter\BlockAccumulator.cs:line 56
   at HandlebarsDotNet.Compiler.BlockAccumulator.ConvertTokens(IEnumerable`1 sequence)+MoveNext() in ..\Handlebars.Net\source\Handlebars\Compiler\Lexer\Converter\BlockAccumulator.cs:line 32
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at HandlebarsDotNet.Compiler.BlockAccumulator.Accumulate(IEnumerable`1 tokens, ICompiledHandlebarsConfiguration configuration) in ..\Handlebars.Net\source\Handlebars\Compiler\Lexer\Converter\BlockAccumulator.cs:line 13
   at HandlebarsDotNet.Compiler.ExpressionBuilder.ConvertTokensToExpressions(IEnumerable`1 tokens, ICompiledHandlebarsConfiguration configuration) in ..\Handlebars.Net\source\Handlebars\Compiler\ExpressionBuilder.cs:line 25
   at HandlebarsDotNet.Compiler.HandlebarsCompiler.Compile(ExtendedStringReader source, CompilationContext compilationContext) in ..\Handlebars.Net\source\Handlebars\Compiler\HandlebarsCompiler.cs:line 25
[...]

PR #607, opened 2025-10-17
Fix GitHub-workflows

TheConstructor:
GitHub auto-cancelled the workflow-runs for #606, so this is an attempt to minimally fix the issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,11 +7,11 @@ on:
 jobs:
   build:
     name: Build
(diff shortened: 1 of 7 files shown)


TheConstructor:
Hey @rexm and @oformaniuk,

is there a chance one of you might have some time to look into this? I guess Sonar is right and the token should be rotated and moved to secrets, but I guess I am unable to help here.

Happy to get some Feedback!

TC

rexm:
This is fixed.

PR #610, opened 2025-11-03
Feature/dotnet8

lennoncork:
Initial work to add .NET 8 as a target


lennoncork:
It seems like it would be good to drop support for netcore3.1, and potentially net4, and net46. Many of the dependent nuget packages no longer support these runtimes. I am happy to do a little work here to clean up but understand that might not be the right call, any guidance appreciated.

lennoncork:
Will merge in changes from #607 when complete.

rexm:
This is now obsolete as we're on 10. Will look at the other PRs.

PR #611, opened 2025-11-06
remove redundant array allocation in HandlebarsCompiler.Compile

SimonCropp:


rexm:
Looks like this change causes some unit test failures

rexm:
## CI failure diagnosis

The benchmark and test failures are caused by removing `.ToArray()` in this PR.

`Tokenizer.Tokenize(source)` returns a lazy `IEnumerable` backed by a forward-only `TextReader`. The old `.ToArray()` eagerly materialised all tokens before any converter touched them. Without it, `BlockAccumulator.ConvertTokens` consumes the enumerator while building block expressions, and a downstream LINQ operation then gets an already-exhausted iterator — causing `System.InvalidOperationException: Sequence contains no elements` in the benchmark and test runs.

The `.ToArray()` is load-bearing: the token stream is a single-pass reader that cannot be re-enumerated. Removing it is not safe here.

**Fix:** restore the `.ToArray()` call, or materialise the sequence at the point where it's first passed into the converter pipeline.

rexm:
Claude recommended the "fix" but I think this PR is just not valid. We do need the array allocation here.

PR #616, opened 2026-06-19
Add Handlebars.js behavioral spec, coverage tests, and net10 dev environment

rexm:
## Summary

- **`HandlebarsSpec.md`** — exhaustive behavioral specification derived from the canonical Handlebars.js spec files (`spec/*.js`), the Mustache spec, official docs, and source analysis. 32 sections covering every feature of the templating language. Includes a new **Known Implementation Gaps** section documenting 5 confirmed behavioral differences from Handlebars.js, each with root cause and backward-compatibility risk rating.

- **`HandlebarsSpecCoverageTests.cs`** — ~100 test methods across 22 behavioral categories covering areas not exercised by the existing suite (HTML encoding edge cases, whitespace control, `#each` data variables, block params, subexpressions, partials, decorators, helper interactions, etc.). Uses the existing `HandlebarsEnvGenerator` pattern so every test runs across 5 `IHandlebars` configurations.

- **Dev environment update** — bumped `Handlebars.Test` and `Handlebars.Benchmark` from EOL `netcoreapp3.1`/`net6` targets to `net10.0`; updated `Microsoft.NET.Test.Sdk` and `BenchmarkDotNet` to current versions.

## Documented implementation gaps (tests assert current behavior as regressions)

| Gap | Spec says | Current behavior | Fix risk |
|-----|-----------|-----------------|----------|
| Default HTML encoder omits `'` `` ` `` `=` | Encode all 7 special chars | `HtmlEncoderLegacy` (default) skips 3; `HtmlEncoder` (opt-in) is correct | HIGH |
| `{{log}}` not registered | No output, calls logger | `HandlebarsRuntimeException` at render time | LOW |
| `false` renders as `"False"` | `"false"` (lowercase) | .NET `bool.ToString()` → `"False"` | MEDIUM |
| Whitespace control on comments | `{{~! x ~}}` strips whitespace | Inline: no strip; block: parse error | LOW |
| `{{^}}` bare caret as else | `{{#x}}a{{^}}b{{/x}}` like `{{else}}` | Resolves to `context.ToString()`; use `{{else}}` | LOW |

## Test plan

- [x] `dotnet test` — 1746/1746 passing (all pre-existing tests plus new coverage tests)
- [x] No library code was changed — all gaps are documented with regression tests only

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
               10.0.x
       - name: Run benchmark
         working-directory: ./source/Handlebars.Benchmark
-        run: dotnet run -c Release --exporters json --filter '*' -m --join
+        run: dotnet run -c Release -f net10.0 --exporters json --filter '*' -m --join
       - name: Get benchmark file name
         id: benchmarkfilename
         working-directory: ./source/Handlebars.Benchmark/BenchmarkDotNet.Artifacts/results
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
           echo $filePath
           echo "file=$filePath" >> $GITHUB_OUTPUT
       - name: Store benchmark result
-        uses: Happypig375/github-action-benchmark@v1.8.2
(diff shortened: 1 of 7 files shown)


rexm:
@oformaniuk we're back baby

oformaniuk (approved):

PR #624, opened 2026-06-20
fix: @partial-block usable in #if and as block partial (issue #519)

### LINKED ISSUE #519: @partial-block not working as partial block or as conditional
### Describe the bug
I'm trying to register a partial that can either be used as a block, with block content, or inline without content. However, both methods I tried using to accomplish this aren't working because `@partial-block` doesn't seem to be properly recognized.  The only way I can get `@partial-block` to render is by using `{{> @partial-block}}` on it's own. Both of the following cases don't render anything:

Block:
```
{{#> @partial-block }}
{{/@partial-block}}
```
Conditional:
```
 {{#if @partial-block}}
  {{> @partial-block}}
  {{/if}}
```

### Expected behavior:
`@partial-block` should be recognized in both the above scenarios and render the block content if it is provided.

### Test to reproduce
```csharp
[Fact]
public void Descriptive_Test_Name_Here()
{
      var template = @"
{{#> myPartial}}
Block content
{{/myPartial}}";
            var partial =
                @"  
Conditional:
  {{#if @partial-block}}
  {{> @partial-block}}
  {{/if}}
Plain:
  {{> @partial-block}}
Block:
{{#> @partial-block }}
{{/@partial-block}}";
            handlebars.RegisterTemplate("myPartial", partial);
            var render = handlebars.Compile(template);
            object data = new { };
            var actual = render(data);
            Assert.Equal("Conditional: Block content Plain: Block content Block: Block content", actual);
}
```

rexm:
Fixes #519

## Summary

- `{{#if @partial-block}}` now correctly evaluates as truthy when the partial is invoked as a block partial (i.e., when a `@partial-block` template is in scope).
- `{{#> @partial-block }}{{/@partial-block}}` no longer crashes with `InvalidOperationException: Sequence contains no elements` when the body between the tags is empty.

## Root causes

**Bug 1 — `{{#if @partial-block}}` always falsy:**
`@partial-block` is a `PathType.Variable`, so its value is resolved via `BindingContext.TryGetContextVariable`. The `PartialBlockTemplate` is stored directly on `BindingContext` and was never exposed through the context data dictionary, so it always returned `UndefinedBindingResult` (falsy). Fixed by adding an explicit check in `TryGetContextVariable` for the segment name `partial-block` that returns the `PartialBlockTemplate` delegate when set.

**Bug 2 — `{{#> @partial-block }}{{/@partial-block}}` crashes:**
`PartialBlockAccumulatorContext.GetAccumulatedBlock()` called `_body.First()` unconditionally, which throws when the body is empty. Fixed by handling the empty-body case (passing `null` as the fallback, which is already the documented no-fallback sentinel in `PartialExpression`).

## Test plan

- [x] New regression test `Issue519_PartialBlockUsableAsBlockAndInIf` added to `IssueTests.cs` — exercises `{{#if @partial-block}}`, `{{> @partial-block}}`, and `{{#> @partial-block}}{{/@partial-block}}` in the same partial.
- [x] Full test suite: 1747 tests, 0 failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/source/Handlebars.Test/IssueTests.cs
+++ b/source/Handlebars.Test/IssueTests.cs
@@ -733,5 +733,22 @@ namespace HandlebarsDotNet.Test
 
             Assert.Throws<HandlebarsCompilerException>(()=> Handlebars.Compile(source));
         }
+
+        // Issue: https://github.com/Handlebars-Net/Handlebars.Net/issues/519
+        [Fact]
+        public void Issue519_PartialBlockUsableAsBlockAndInIf()
+        {
+            var handlebars = Handlebars.Create();
+            handlebars.RegisterTemplate("myPartial",
+                @"Conditional:{{#if @partial-block}} {{> @partial-block}}{{/if}}
+Plain: {{> @partial-block}}
+Block:{{#> @partial-block }}{{/@partial-block}}");
+
+            var render = handlebars.Compile("{{#> myPartial}}Block content{{/myPartial}}");
+            var actual = render(new { });
+            Assert.Contains("Conditional: Block content", actual);
+            Assert.Contains("Plain: Block content", actual);
+            Assert.Contains("Block:Block content", actual);
+        }
     }
 }
--- a/source/Handlebars/BindingContext.cs
+++ b/source/Handlebars/BindingContext.cs
@@ -156,7 +156,20 @@ namespace HandlebarsDotNet
                 return BlockParamsObject.TryGetValue(segment, out value)
                    || ContextDataObject.TryGetValue(wellKnownVariable, out value);
             }
-            
+
+            // Special handling for @partial-block: return the PartialBlockTemplate delegate
+            // so that {{#if @partial-block}} is truthy when a block partial is provided.
+            if (segment.LowerInvariant == "partial-block")
+            {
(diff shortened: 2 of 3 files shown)


source/Handlebars.Test/IssueTests.cs
```diff
@@ -734,6 +734,23 @@ public void UnrecognisedExpressionThrowsOutOfMemoryException()
...
 
+        // Issue: https://github.com/Handlebars-Net/Handlebars.Net/issues/519
+        [Fact]
+        public void Issue519_PartialBlockUsableAsBlockAndInIf()
+        {
+            var handlebars = Handlebars.Create();
+            handlebars.RegisterTemplate("myPartial",
+                @"Conditional:{{#if @partial-block}} {{> @partial-block}}{{/if}}
+Plain: {{> @partial-block}}
+Block:{{#> @partial-block }}{{/@partial-block}}");
```
TheConstructor, line 745: Using
```
Block:{{#> @partial-block }}any fallback{{/@partial-block}}
```

should still let the test pass, but doesn't, right? I am a bit stuck in figuring out the right solution for this, so I can continue with #606

PR #642, opened 2026-06-23
Nullable Reference Types

TheConstructor:
I tried my best to annotate everything with Nullable Reference Annotations, and to convince the compiler of them (i.e. move assignments to the expected positions, ...). It's definitely a big PR. Sorry.

### DIFF
--- a/source/Handlebars.Benchmark/Compilation.cs
+++ b/source/Handlebars.Benchmark/Compilation.cs
@@ -6,7 +6,7 @@ namespace HandlebarsNet.Benchmark
 {
     public class Compilation
     {
-        private IHandlebars _handlebars;
+        private IHandlebars _handlebars = null!;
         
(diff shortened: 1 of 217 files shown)


rexm:
What is the purpose?

TheConstructor:
> What is the purpose?

It may make it more obvious to the user of the library when Handlebars.Net expect something not to be `null` on one hand, and it will make the compiler push you to make `null`-checks at the right times - at least in theory. There are some situations, were the tooling isn't advanced enough to understand when something can or can't be `null`. Still it usually leads to fewer unthought of NullReferenceExceptions.

If you are new to NRT you might read https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis - it contains some use-cases and how describes how the tooling tries to help you

